Issue #24, opened 2019-02-24
Upgrade `urql` devDependency and peerDependency.


parkerziegler:
As the first step to starting work on `v1`, we'll need to upgrade the `urql` devDependency and peerDependency to `^1.0.0`.

parkerziegler:
Closed by #25.

PR #25, opened 2019-02-24
#24 Upgrade urql devDependency and peerDependency.

parkerziegler:
This PR upgrades our devDependency and peerDependency on `urql` to `^1.0.0` and closes issue #24. We can now begin safely targeting the v1 branch with the correct version of `urql`. Since v1 publishes of `urql` are coming in hot right now, we'll want to make sure we update our `yarn.lock` every so often to make sure we have latest. We could also probably benefit from a clean `yarn.lock` to ensure we have the latest deps more generally (esp. `bs-platfotm`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "react-dom": "^16.2.0",
     "reason-react": ">=0.3.4",
     "url-loader": "^1.1.2",
-    "urql": "^0.2.1",
+    "urql": "^1.0.0",
     "webpack": "^4.0.1",
     "webpack-cli": "^3.1.1",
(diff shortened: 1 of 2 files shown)


parkerziegler:
Also, @kiraarghy and @kitten, this predictably does break ~75% of our breaks. I'm personally ok with merging branches that fail into `v1` so that we can move a bit faster, and just ensure tests are passing and snapshots get updated when we PR `v1` into `master`. How do you two feel about that strategy?

kiraarghy:
@parkerziegler agree for development on v1 branch, failing tests is fine!

Issue #26, opened 2019-02-27
Update bindings for `Client`.


parkerziegler:
As part of the `v1` migration, we'll need to update the bindings for `Client`. Probably the most significant change will be the addition of bindings for [exchanges](https://github.com/FormidableLabs/urql/blob/274fd4854bb80aa9baeb71ea6935cc06a6580631/src/types.ts#L52). This may be simplified by the fact that, under the hood, exchanges are powered by [ `wonka`](https://github.com/kitten/wonka), whose source is written in Reason. Other than that, our life is greatly simplified b/c we no longer have cache operations to keep track of in the `Client`.

parkerziegler:
Roadmap:

- [x] Add support for `url` and `fetchOptions`, with proper unwrapping of `fetchOptions` variant.
- [x] Add support and example for `executeQuery` and `executeMutation`.
- [ ] Add support for shipped exchanges.
- [ ] Add support and example for using some of the shipped exchanges.
- [ ] Add support for custom exchanges.

PR #27, opened 2019-03-03
#26 Update bindings for Client.

parkerziegler:
This PR partially addresses issue #26. The changes to the client in `urql` v1 are substantial enough that this change will likely need to happen piece by piece. The pieces I've included here:

- Add a top level `UrqlTypes` module. This is where we should share types used by multiple modules. I added a type for `graphqlRequest` and also began to scaffold out how we might bind the `requestPolicy` enum using polymorphic variants a la: https://bucklescript.github.io/docs/en/generate-converters-accessors#convert-between-js-string-enum-and-bs-polymorphic-variant
- Added the `UrqlRequest` module, which just holds the `createRequest` function. This generates valid `graphqlRequest`s for us, and should be used by consumers using these APIs directly.
- Updated the client bindings to remove all the custom cache stuff from previous versions.
- Updated bindings for `executeQuery` and `executeMutation`. Since these operations now return Wonka sources, we have first class Reason types 🎉. We'll be able to type these bindings even further once we type `OperationResult`: https://github.com/FormidableLabs/urql/blob/774e09db8613d79e3e5cc6fca3d7f91526765f8e/src/types.ts#L44
- Added a new `examples` dir. The old example will need to be mostly thrown out, since much of it is invalid w/ the new `urql`. I'm keeping it around for help in writing new examples now, but it no longer gets compiled by `bsb` since I've removed it from our `bsconfig.json`. 
- Added a basic example showing `executeQuery` and `executeMutation` in action w/ Wonka.

@kiraarghy @kitten Hopefully this can help kickstart some other work / get folks unstuck. I realized that starting w/ binding the components and hooks felt a bit difficult w/ out a proper client, so I wanted to get this up ASAP. Furthermore, I'll be totally offline from now until next Sunday visiting some fam, so feel free to merge this if you want to get these changes into your local.


__tests__/UrqlClient_test.re
```diff
@@ -1,29 +1,21 @@
 open Jest;
 open ReasonUrql;
 
+let it = test;
```
parkerziegler: Happily, these few tests pass!

bsconfig.json
```diff
@@ -9,7 +9,7 @@
       "subdirs": true
     },
     {
-      "dir": "example",
+      "dir": "examples",
```
parkerziegler: Removing the `example` dir completely from compilation, since it'll totally fail the compiler w/ the new bindings.

bsconfig.json
```diff
@@ -27,7 +27,7 @@
   ],
   "suffix": ".bs.js",
   "namespace": false,
-  "bs-dependencies": ["reason-react", "bs-fetch"],
+  "bs-dependencies": ["reason-react", "bs-fetch", "wonka"],
```
parkerziegler: Make sure to build `wonka` so we can use it. @kitten do you recommend we take an explicit dep on `wonka` and hope it gets deduped, or just reference the one brought in by `urql`?

src/UrqlClient.re
```diff
@@ -1,97 +1,40 @@
...
-
-[@bs.new] [@bs.module "urql"]
-external createClient: clientConfig('store, 'value) => client = "Client";
+[@bs.new] [@bs.module "urql"] external client: clientOptions => t = "Client";
 
 [@bs.send]
 external executeQuery:
-  (~client: client, ~query: UrqlQuery.urqlQuery, ~skipCache: bool) =>
-  Js.Promise.t('a) =
+  (~client: t, ~query: UrqlTypes.graphqlRequest) => Wonka.Types.sourceT('a) =
```
parkerziegler: This type param to `sourceT` here should be an `OperationResult` as defined here: https://github.com/FormidableLabs/urql/blob/774e09db8613d79e3e5cc6fca3d7f91526765f8e/src/types.ts#L44.

webpack.config.js
```diff
@@ -1,16 +1,16 @@
 const path = require("path");
 
 module.exports = {
-  entry: path.join(__dirname, "example/app/Example.bs.js"),
+  entry: path.join(__dirname, "examples/1-basic/app/Basic.bs.js"),
```
parkerziegler: Eventually we'll want to have this `webpack` setup in each example dir and have them isolated, just like `urql`.

bsconfig.json
kitten: Just take in the explicit dependency :+1: There's no peer dependency on it, but ultimately we only have to make sure that our version ranges match major versions.

Issue #28, opened 2019-03-14
Add `examples` directory using yarn workspaces.


parkerziegler:
Similar to how `urql` has an `examples` directory (modeled loosely after Next's), we'll want to have some `examples` to demonstrate how to use `reason-urql` with ReasonReact. We should use yarn workspaces to manage dependencies across individual examples. Each should therefore have their own `package.json` and `bsconfig.json`. For the most part, all dependencies can be shared (`bs-platform`, `webpack`, `bs-css`, `reason-react`) and we'll take a `file:` dependency on `reason-urql`.

parkerziegler:
### Roadmap

- [X] Add directory and setup yarn workspaces.
- [X] Add `reason-urql` as `file` dependency.
- [ ] Fix duplicated dependency issue caused by dependency on `reason-urql` 🤦‍♂️

parkerziegler:
Closed by #29.

PR #29, opened 2019-03-14
#28 Add examples directory with yarn workspaces, bindings for Mutation, Provider.

parkerziegler:
This PR begins to address #28. To make development of new features easier, it's nice to have an easy to work with set of `examples` to work with. This adds an `examples` directory managed by yarn workspaces. To run the examples a user need only:

1. `cd examples`
2. `yarn`
3. cd `1-execute-query-mutation` (or whichever one they want to run).
4. `yarn start` in one terminal to compile the source, `yarn start:demo` to start `webpack` (with hot reloading)!

Here's what running `1-execute-query-mutation` looks like currently:

ExecuteQueryMutation

It's just a fun little previewer that lets you see data updating across queries and mutations.

This PR **also** adds initial bindings for the `Mutation` component. The coolest part of these bindings is the use of `urqlDataToRecord`, which converts the Js.t we receive from the JS side into a record with a special `response` key. `response` is a variant of the shape:

```reason
type response('a) =
| Fetching
| Data('a)
| Error(UrqlError.t)
| NotFound
```

This allows us to do some _really_ nice pattern matching with ReasonReact, and is in accordance with `reason-urql` v0.1.1. See the `2-mutation` example (which currently does nothing 😛, want to get `Query` in before fleshing out that example) to check this out.

Finally, there are bindings for `Provider` so that we can wrap our app roots and actually use `Mutation` and `Query` 🎉


src/components/UrqlMutation.re
```diff
@@ -0,0 +1,55 @@
+[@bs.module "urql"] external mutation: ReasonReact.reactClass = "Mutation";
+let component = ReasonReact.statelessComponent("Mutation");
+
+type mutationRenderPropsJs('a) = {
+  .
+  "fetching": bool,
+  "data": Js.Nullable.t('a),
+  "error": Js.Nullable.t(UrqlError.t),
+  "executeMutation": Js.Json.t => unit,
```
parkerziegler: In reviewing the `urql` docs, I think this actually returns a `Promise`. cc/ @kitten

src/utils/UrqlError.re
```diff
@@ -0,0 +1,10 @@
+type t;
+
+[@bs.deriving abstract]
+type error = {
+  [@bs.optional]
+  networkError: Js.Exn.t,
+};
+
+[@bs.new] [@bs.module "urql"]
+external combinedError: string => t = "CombinedError";
```
parkerziegler: This is just total scaffolding to get it in the right place, but this needs specific work!

src/components/UrqlMutation.re
kitten: That's correct 👌

Issue #22, opened 2019-02-19
Add bindings for `Query` component.


parkerziegler:
`urql` v1 adds a specific component for `Query`, in addition to a `useQuery` hook: https://github.com/FormidableLabs/urql/blob/master/docs/api.md.

For this issue, it'd be ideal to just focus on binding these two pieces of the API. Binding `Query` should be as simple as using `ReasonReact.wrapJsForReason` and typing the props. We will want to re-use some of the nice utilities we have in place to convert `fetching`, `data`, `error` to a variant (see [`urqlDataToVariant`](https://github.com/FormidableLabs/reason-urql/blob/master/src/UrqlConnect.re#L62-L83)). We should keep that function generic so we can 1) unit test it, and 2) use it for `Mutation`. Binding the hook should also be fairly easy using standard BuckleScript FFI for functions.

PRs for this issue should target the `v1` branch.

kiraarghy:
I'll take this! 

![](https://media.giphy.com/media/zaezT79s3Ng7C/giphy.gif)

parkerziegler:
Dope! Assigning you now. I'll probably take `Mutation` so we can each take a stab at our own approaches and see where we land relative to one another.

parkerziegler:
FYI @kiraarghy let's copy the folder structure from `urql` exactly. So for the `Query` component, let's put it in `src/components/UrqlQuery.re` and for the hook let's put it in `src/hooks/UseQuery.re` or something of that sort. Why the weird naming for `UrqlQuery`? Well, one thing I've thought about is that consumers of `reason-urql` might have their own `Query` module somewhere in the tree, with which a `Query` module might collide. So by namespacing it: http://2ality.com/2017/12/modules-reasonml.html#namespacing-modules, we prevent these potential collisions. In our top level `ReasonUrql.re` we'll just alias them to their true names, i.e.

```reason
/* ReasonUrql.re */
module Query = UrqlQuery
```

I'd also be curious about using snake case for modules rather than pascal case. I know Wonka uses snake case 🤔 Any feelings or opinions here? What reads to you as more Reason-y.

kiraarghy:
@parkerziegler 
* Agree with the folder structure and namespacing, thanks for explaining!
* I generally prefer pascal case cause snake case brings up memories of my bad python code 🐍. It feels more Reason/OCaml-esque TBH.

parkerziegler:
Cool! Let's stick with pascal case! 🚀 🚫 🐍

[...]

PR #31, opened 2019-03-20
#22 Implements Query Component Bindings.

kiraarghy:
This PR implements the urql v1.0.0 Query component bindings. 

Any feedback on API design/ documentation is welcome! 

ToDo: 

1. @parkerziegler and I have talked about the potential for testing, will be looking into writing bs-enzyme/bs-jest tests for this component.

2. Additional documentation where it makes sense.

3. As a separate PR produce a nice example documenting the use of the Query component in ReasonReact.


src/components/UrqlQueryNew.re
```diff
@@ -0,0 +1,67 @@
...
+  variables: Js.Json.t,
+  client: Client.t,
+  requestPolicy: string,
+};
+
+let make = (~query, ~client, requestPolicy, children) => {
+  let typedRequestPolicy = requestPolicyGetter(requestPolicy);
+  ReasonReact.wrapJsForReason(
+    ~reactClass=queryComponent,
+    ~props=jsProps(~query, ~client, ~requestPolicy=typedRequestPolicy),
```
parkerziegler: I think we want `~variables` here rather than `~client` 😄

src/utils/RequestPolicyUtil.re
```diff
@@ -0,0 +1,20 @@
+type requestPolicy =
```
parkerziegler: This looks really sweet! Looking over this API, I'm still tinkering with a few things:

1. I'm wondering if we can just use the power of default props to take away the need for `requestPolicySet` and `requestPolicyGetter`. It seems like we could have a default set in `make` function such that, if a user doesn't set this, we'll just pass it in for them.
2. I began to work on binding `requestPolicy` using `[@bs.deriving jsConverter]` with `[@bs.as]`: https://github.com/FormidableLabs/reason-urql/blob/v1/src/UrqlTypes.re#L1-L7 This construct is a special type of Bucklescript annotation specifically for string enums like this: https://bucklescript.github.io/docs/en/generate-converters-accessors#convert-between-js-string-enum-and-bs-polymorphic-variant.

Combining the above two together I'm envisioning something like the following in `UrqlQuery.re`:

```reason
let make = (~query, ~variables, ~requestPolicy=`CacheFirst, children) => {
  /* yada yada yada */
  ~props=jsProps(~query, ~variables, ~requestPolicy=requestPolicyToJs(requestPolicy)),
}
```

I _think_ that would convert whatever polymorphic variant the user passes in for `requestPolicy` into the appropriate JS string while still maintaining type validation. What do you think? The last thing I'll add is that I like the idea of this living in `UrqlTypes.re` since we'll be needing to reference it in other parts of the API like `OperationContext`.

src/components/UrqlQueryNew.re
```diff
@@ -0,0 +1,67 @@
...
+
+type queryRenderPropsJs('a) = {
+  .
+  "fetching": bool,
+  "data": Js.Nullable.t('a),
+  "error": Js.Nullable.t(UrqlError.t),
+  "executeQuery": Js.Json.t => Js.Promise.t('a),
+};
+
+type response('a) =
```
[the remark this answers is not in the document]
kiraarghy: Yep! Totally agree, will do this in my next commit!

src/components/UrqlQueryNew.re
```diff
@@ -0,0 +1,67 @@
+open ReasonUrql;
+open RequestPolicyUtil;
+
+[@bs.module]
+external queryComponent: ReasonReact.reactClass = "urql/src/components/Query";
```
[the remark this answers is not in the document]
kiraarghy: I did not know you could do this, extremely cool!

src/components/UrqlQueryNew.re
```diff
@@ -0,0 +1,67 @@
...
+    response,
+  };
+};
+
+[@bs.deriving abstract]
+type jsProps = {
+  query: string,
+  [@bs.optional]
+  variables: Js.Json.t,
+  client: Client.t,
```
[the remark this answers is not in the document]
kiraarghy: Yep!

src/utils/RequestPolicyUtil.re
kiraarghy: I didn't know you could use default props, have included it! The api seems a lot more terse than my solution 😄!

kiraarghy:
@parkerziegler how does the update PR look like? 😄

src/components/UrqlQuery.re
```diff
@@ -0,0 +1,61 @@
...
+    data,
+    error,
+    executeQuery: result##executeQuery,
+    response,
+  };
+};
+
+[@bs.deriving abstract]
+type jsProps = {
+  gqlTemplate: string,
```
kiraarghy: I don't really like this but am getting namespacing issues between the classname being 'query' and the query prop being named 'query'. Any suggestions @parkerziegler?

parkerziegler: Ah, I hadn't thought about that. Since the React class `external` is only bound to this module and never exposed externally, let's fallback to your original nomenclature of `queryComponent` and stick with keeping the prop named as `query`. I'll update `mutation` to `mutationComponent` after we merge this to keep it consistent.

parkerziegler:
Looks great @kiraarghy! If you're cool with it, can we rebase down to a single commit? Then let's get this puppy merged! Awesome work – I'll push up a PR to get you on the all contributors list tomorrow 🎉 💯!

Issue #32, opened 2019-03-23
Add bindings for `CombinedError`.


parkerziegler:
We need to bind the `CombinedError` that `urql` exposes and that gets passed along to the render props for `Query`, `Mutation`, and `Subscription`. https://github.com/FormidableLabs/urql/blob/master/docs/api.md#combinederror-class

parkerziegler:
Closed by #33.

PR #33, opened 2019-03-23
#32 Add CombinedError. Clean up dependency tree.

parkerziegler:
This PR addresses #32 by adding bindings for `CombinedError` to `reason-urql` 🎉

Interestingly, it is rather difficult to produce a case where a `GraphQLError` gets produced because of the compile-time query validation provided by `graphql_ppx`. I'm also attempting to uncover a way to test the bindings for network errors, but also having difficulty creating those conditions. Even when I block requests in Chrome dev tools I'm not seeing an error propagate down to `reason-urql`'s render props 🤔cc/ @kiraarghy @kitten would love any suggestions you have on actually seeing these bindings work in the wild 😅.

I also cleaned up the dependency tree quite a bit to remove unnecessary dev dependencies. Most were leftover relics from when we kept our example GraphQL server in source.


kiraarghy:
Re: the Errors, I have no idea how to simulate it! Phil is probably the expert?

Issue #36, opened 2019-04-02
Implement CI pipeline for reason-urql.


kiraarghy:
It would be nice to start implementing a CI pipeline for reason-urql post V1 to ensure that we don't have build issues, or an Urql update causes our bindings to fail. 

@kadikraman seems to be having success using Sail CI for [React Native App Auth](https://github.com/FormidableLabs/react-native-app-auth) so I'm up for using that! 

Goal: Implement a simple build and verification pipeline for reason-urql with the goal to expand it later on!

parkerziegler:
Closed by #37, thanks @kiraarghy!

PR #37, opened 2019-04-02
#36 implement Sail CI

kiraarghy:
As per issue, created config for Sail CI that installs and builds reason-urql. 

When/ if we decide to write tests for v1 we can then add a test stage.

Pretty simple IMO @parkerziegler any thoughts? 

ToDo: Add Sail CI app to project if we decide to go with it.

### DIFF
--- a/.sail.yml
+++ b/.sail.yml
@@ -1,0 +1,19 @@
+tasks:
+  install:
+    image: "node:carbon"
+    command:
+      - yarn
+    args:
+      - install
+  build:
+    image: "node:carbon"
+    command:
+      - yarn
+    args:
+      - build
+  test:
+    image: "node:carbon"
+    command: 
+      - yarn
+    args:
+      - test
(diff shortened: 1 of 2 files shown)


parkerziegler:
@kiraarghy noticed that you closed this. Any particular reason? Definitely happy to give Sail a shot! When we merge this, we should also remove the `.travis.yml` so we're not running CI in two places 😄 . Is there a dashboard where we can see CI runs? Also, go ahead and add the test script, we do have a few tests for the client right now on `v1`!

parkerziegler:
@all-contributors please add @kiraarghy for code and test.

kadikraman:
@parkerziegler - they use GitHub Checks, but they've also got a UI here, which you only really need if you want to add env vars - https://app.sail.ci 

I used it for [react-native-app-auth](https://github.com/FormidableLabs/react-native-app-auth) (for testing) and [threadstarter](https://github.com/kadikraman/threadstarter) (for testing and deployment to AWS).

PR #39, opened 2019-04-05
Remove Travis CI and Coveralls. Add Codecov as coverage reporting tool.

parkerziegler:
This PR removes Travis CI as our CI provider, adds Codecov as our coverage reporting tool, and modifies our `.sail.yml` to work with environment variables. There are still some kinks to work out with Codecov – right now we aren't actually getting reporting showing up in their dashboards. I'm suspecting some of this is because we're introducing this change on a branch off a branch off of `master`. There are some issues with Codecov not having a "base report" that I've attempted to amend via by using their [bash uploader](https://docs.codecov.io/v4.3.0/docs/about-the-codecov-bash-uploader), but it's not quite behaving properly. I have a support ticket out to them now that we'll hopefully get some word on soon. In the meantime, our Sail builds are passing nicely 👍.


.sail.yml
```diff
@@ -13,7 +13,22 @@ tasks:
...
+      - coverage
+  codecov:
+    image: "node:carbon"
+    env:
+      - CODECOV_TOKEN
+    command:
+      - yarn
+    args:
+      - codecov
+      - --token=$(CODECOV_TOKEN)
```
ryan-roemer: Isn’t this supposed to be ${} and not $() or is it some domain specific interpolation that’s not shell?

ryan-roemer:
Might want to review inspectpack which also does codecov but without any token vars

.sail.yml
parkerziegler: Yep, that's originally what I thought too. Then I read Sail's docs on environment variables: https://sail.ci/docs/environment-variables. Check out the "Use Environment Variables in Task Properties" section, apparently they do interpolation using `$()`.

parkerziegler:
Yeah, I took a look at both `inspectpack` and `react-ssr-prepass`. It looks like Codecov can auto-detect those Circle and Travis, so a token isn't necessary when using those CIs. What is stumping me is that I tried running this build with Travis still enabled: https://travis-ci.org/FormidableLabs/reason-urql/builds/516009533. It reaches the `codecov` step just fine and even outputs a success state in the build log. However, if you follow [the link to the Codecov report](https://codecov.io/github/FormidableLabs/reason-urql/commit/fff761d72899e213e43f020ef8bd29a15e0cff55), it responds with an opaque "There was an error processing coverage reports." I've checked the common issues, and the only one I think it _might_ be are the coverage reports not resolving, but I can't find a way to debug that with Codecov.

parkerziegler:
image

This is probably what confuses me the most – the "Missing Base Report" notification. I may try moving the default branch to `v1` for now and seeing if using the bash uploader to upload a report to `v1` helps generate the proper report. Like I mentioned, I have an email out to Codecov and will hopefully get some help from them.

.sail.yml
ryan-roemer: Interesting! So, does that mean they're pre-processing the interpolation to be something like:

```sh
$ export CODECOV_TOKEN="hi"
$ yarn codecov --token=hi
```

?

Because if this was _bash_ with `$(STRING)` the `STRING` would be executed as a command. Curious if doing `- --token=${CODECOV_TOKEN}` produces

```sh
$ export CODECOV_TOKEN="hi"
$ yarn codecov --token=${CODECOV_TOKEN}
```

.. which seems like it should work too...

parkerziegler: My guess is they must be doing something like the above. I believe I ran an experiment where I had the CI run `echo ${CODECOV_TOKEN}` and it worked. The key to all of this is ensuring that the `env` key on the task is specified so that the environment variable you want to use is mounted: https://github.com/FormidableLabs/reason-urql/blob/v1/.sail.yml#L28-L29. Basically, the steps seem to be:

- Add env variables through Sail's app (app.sail.ci).
- Specify which variables to bring into scope in which task (using the `env` config option as shown above).
- Interpolate the variable using `$()`.

Sail does have a few other drawbacks from my research thus far.

- You can't see the build logs in their UI (app.sail.ci). The only way to see them is via the "Checks" tab on a PR. Which means you just don't get logs for non-PR'ed builds.
- Their environment variables UI seems badly broken. You can't actually save environment variables without opening the dev tools and bringing the Save button into view (or praying you tab navigate to the right out of view button 😬). 
SailUI
- Their deletion of environment variables doesn't seem to be working. If you delete variables in the UI and then refresh the page, your variables will still appear (and work in CI, meaning they're being persisted).
SailNoDeletion

The CI itself seems to be running fine, but in the longer run I probably will advocate for us to migrate to Circle if these issues persist.

Issue #40, opened 2019-04-07
Add bindings for Exchanges.


parkerziegler:
Previously this issue was tracked in #26, but it's complex enough that I figured we'd make it a separate issue. We want `reason-urql` to add support for [exchanges](https://github.com/FormidableLabs/urql/blob/master/docs/basics.md). This shouldn't be too difficult, as `urql` already has a set of exchanges that are a part of the core API. We just need to expose those using `[@bs.module]` declarations. Adding exchange types will be a bit more complex and should be based off of the [`urql/types.ts` module](https://github.com/FormidableLabs/urql/blob/master/src/types.ts).

### Roadmap
- [x] Add bindings for exchange types.
- [x] Add support for shipped exchanges.
- [x] Add support and example for using some of the shipped exchanges.
- [x] Add support for custom exchanges. 
- [x] Add specific config for `subscriptionExchange`.

parkerziegler:
Closed by #41 and #42.

PR #41, opened 2019-04-07
#40 Add support for exchanges.

parkerziegler:
This PR adds support for exchanges 🎉, addressing #40. The bulk of the work went into typing all of the API surface that underpins exchanges, i.e. `operation`, `operationResult`, `operationType`, `operationContext`, `exchangeInput`, and `exchangeIO`, but that was made much easier by `urql`'s TS defs.

A note on the architecture here – you'll see that `UrqlExchanges` is a `module` defined within `UrqlClient`. There's a reason for this. Exchanges accept `exchangeInput` as their single parameter, which is an object composed of `forward` (a function for forwarding operations to the next exchange) and `client` (the client instance, in this case of type `UrqlClient.t`). Because of this pairing between the types for exchanges and the type `t` of the client, the modules have to be co-located or we get the cyclic dependency of death error.

For example, if `UrqlExchanges` was its own module our dependency graph would look like:

- `UrqlExchanges.exchangeInput` references `UrqlClient.t`.
- `UrqlClient` references `UrqlExchanges.exchange` as part of its config options.
- Cyclic dependency error :rip:

Co-locating _makes sense_ here though, because exchanges are really only used by the client. If other folks have other organizational strategies, I'd love to hear them!

### DIFF
--- a/examples/1-execute-query-mutation/src/Example.re
+++ b/examples/1-execute-query-mutation/src/Example.re
@@ -91,7 +91,7 @@ module Styles = {
       maxHeight(px(200)),
       overflow(`auto),
       minWidth(px(300)),
-      maxWidth(px(500))
+      maxWidth(px(500)),
     ]);
 };
 
--- a/examples/4-exchanges/src/index.re
+++ b/examples/4-exchanges/src/index.re
@@ -1,9 +1,27 @@
 open ReasonUrql;
 
+/* This is the native debugExchange that ships with `urql`, re-implemented in Reason.
+     Typically, you'd just add Exchanges.debugExchange to the Client's exchange array.
+   */
+let debugExchange: Exchanges.exchange =
+  exchangeInput => {
+    let forward = Exchanges.forwardGet(exchangeInput);
+
+    ops =>
+      ops
+      |> Wonka.tap((. op) =>
+           Js.log2("[Exchange debug]: Incoming operation: ", op)
+         )
+      |> forward
+      |> Wonka.tap((. res) =>
+           Js.log2("[Exchange debug]: Completed operation: ", res)
+         );
+  };
+
 let client =
   Client.make(
     ~url="https://formidadog-ql.now.sh",
-    ~exchanges=[|Exchanges.debugExchange, Exchanges.fetchExchange|],
+    ~exchanges=[|debugExchange, Exchanges.fetchExchange|],
     (),
   );
 
@@ -11,6 +29,7 @@ module GetAllDogs = [%graphql
   {|
   query dogs {
     dogs {
+      key
       name
       breed
       likes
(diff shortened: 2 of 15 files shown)


examples/1-execute-query-mutation/src/Example.re
```diff
@@ -91,7 +91,7 @@ module Styles = {
       maxHeight(px(200)),
       overflow(`auto),
       minWidth(px(300)),
-      maxWidth(px(500))
```
parkerziegler: I'm going to add a `refmt` pre-commit hook in soon to catch things like this!

examples/4-exchanges/src/index.re
```diff
@@ -0,0 +1,78 @@
...
+    likeDog(key: $key) {
+      name
+      likes
+    }
+  }
+  |}
+];
+
+Client.executeQuery(~client, ~query=queryRequest)
+|> Wonka.subscribe((. response) => {
```
parkerziegler: I'll flesh this example out more a bit later on. Right now, it just kicks off a mutation every 5 seconds so you can see the `debugExchange` working in the console.

package.json
```diff
@@ -22,24 +22,21 @@
...
+    "urql": "^1.0.4"
   },
   "devDependencies": {
     "@glennsl/bs-jest": "^0.4.8",
     "all-contributors-cli": "^5.4.0",
     "bs-platform": "^4.0.18",
-    "codecov": "^3.3.0",
-    "coveralls": "^3.0.2",
-    "graphql": "^14.1.1",
+    "codecov": "3.2.0",
```
parkerziegler: Still messing around trying to get Codecov working. I think it might be too messed up on the current project, so we'll retry with a new token when `v1` is in a mergeable state.

src/UrqlClient.re
```diff
@@ -9,11 +9,73 @@ let unwrapFetchOptions = fetchOptions =>
...
+  };
+
+  type exchange = exchangeInput => exchangeIO;
+
+  [@bs.module "urql"] external cacheExchange: exchange = "";
+  [@bs.module "urql"] external debugExchange: exchange = "";
+  [@bs.module "urql"] external dedupExchange: exchange = "";
+  [@bs.module "urql"] external fallbackExchangeIO: exchangeIO = "";
+  [@bs.module "urql"] external fetchExchange: exchange = "";
+  [@bs.module "urql"] external subscriptionExchange: exchange = "";
```
parkerziegler: Looking more closely, `subscriptionExchange` may differ slightly in API from the other exchanges. I'll circle back to these bindings in the next PR, which'll add tests!

kitten: Yea that is a factory that takes some extra options

PR #42, opened 2019-04-12
Add bindings for subscription exchange.

parkerziegler:
This PR adds specific bindings for `subscriptionExchange`. This exchange is a bit more complex than the others, accepting a `SubscriptionForwarder` and doing some transformations to our subscriptions under the hood. While I don't _fully_ understand how this works, I'm mostly following the interfaces outlined [here](https://github.com/FormidableLabs/urql/blob/0dfa8954c46d0b10d298805ffdad9da5c0d231f9/src/exchanges/subscription.ts#L24-L62). @kitten would appreciate your insight here as much as possible. I'll  try to get moving on implementing an example similar to @andyrichardson's to see if these bindings are halfway decent.

### DIFF
--- a/__tests__/UrqlClient_test.re
+++ b/__tests__/UrqlClient_test.re
@@ -54,4 +54,17 @@ describe("UrqlClient", () => {
       Expect.(expect(client) |> toMatchSnapshot);
     });
   });
+
+  describe("Client with exchanges provided", () =>
+    it("should instantiate a client with exchanges", () => {
+      let client =
+        Client.make(
+          ~url="https://localhost:3000",
+          ~exchanges=[|Exchanges.debugExchange|],
+          (),
+        );
+
+      Expect.(expect(client) |> toMatchSnapshot);
+    })
+  );
 });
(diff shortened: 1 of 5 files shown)


__tests__/UrqlClient_test.re
```diff
@@ -54,4 +54,17 @@ describe("UrqlClient", () => {
       Expect.(expect(client) |> toMatchSnapshot);
     });
   });
+
+  describe("Client with exchanges provided", () =>
```
parkerziegler: Just verifying that the `Client` can actually accept an exchanges array and pass the compiler.

src/utils/UrqlCombinedError.re
```diff
@@ -34,12 +34,12 @@ type positions = array(int);
 type originalError = Js.Exn.t;
 
 /* Extension fields to add to the formatted error. */
-type extensions('a) = Js.t({..} as 'a);
+type extension;
```
parkerziegler: Decided to make this type abstract. Basically, an index signature like this can be implemented in BuckleScript using `Js.Dict.t`, but the type of `extension` is `any` in the official `graphql` bindings, so it seems to make sense to leave it abstract here.

kitten: That seems like the right choice; It's impossible to know what a user might stick in here. I'd say the only "safe" choice would be a `Js.Json`, which implies to a user of these bindings that `bs-json` et al can be used to decode the extensions

Issue #35, opened 2019-03-25
Add bindings for `Subscription`.


parkerziegler:
Now that we have bindings for both `Query` and `Mutation` in place, let's complete the loop with `Subscription`! @kiraarghy assigning this to you for now, but feel free to pick it up only when you have time.

parkerziegler:
FYI @kiraarghy I had some time this afternoon and took these on! Will re-assign the issue over to me for the time being. Want to make sure everything with `subscriptionExchange` is working and that's a bit tough without a `Subscription` component.

parkerziegler:
Closed by #47.

PR #47, opened 2019-04-20
#35 Add bindings for Subscription.

parkerziegler:
This PR adds bindings for the `Subscription` component. These mostly follow the conventions outlined in `Query` and `Mutation`, with a few specifics that occur only on the `Subscription` component in `urql`. A few things to note:

1. The type for `handler` (`type handler('a, 'b) = ('a, 'b) => 'a`). Basically, `handler` receives the previous subscriptions and the current subscription, and returns the same structure as previous subscriptions. So type parameters are a great way to handle this. I may consider making these named arguments as well for clarity if folks agree.
2. The pattern matching in the `make` function of `Subscription`. Basically, we have two optional props, so we pattern match on a tuple of those two props (`variables` and `handler`) to determine which props to pass to `Subscription`. `urql`'s defaults help us out here and make this nicely ergonomic for Reason consumers.

I'm working on an example right now which will come up in a separate PR.

### DIFF
--- a/src/components/UrqlMutation.re
+++ b/src/components/UrqlMutation.re
@@ -6,14 +6,16 @@ type mutationRenderPropsJs('a) = {
   "fetching": bool,
   "data": Js.Nullable.t('a),
   "error": Js.Nullable.t(UrqlCombinedError.t),
-  "executeMutation": Js.Json.t => Js.Promise.t('a),
+  "executeMutation":
+    Js.Json.t => Js.Promise.t(UrqlClient.UrqlExchanges.operationResult),
 };
 
 type mutationRenderProps('a) = {
   fetching: bool,
   data: option('a),
   error: option(UrqlCombinedError.t),
-  executeMutation: Js.Json.t => Js.Promise.t('a),
+  executeMutation:
+    Js.Json.t => Js.Promise.t(UrqlClient.UrqlExchanges.operationResult),
   response: UrqlTypes.response('a),
 };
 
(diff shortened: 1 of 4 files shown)


src/components/UrqlMutation.re
```diff
@@ -6,14 +6,16 @@ type mutationRenderPropsJs('a) = {
   "fetching": bool,
   "data": Js.Nullable.t('a),
   "error": Js.Nullable.t(UrqlCombinedError.t),
-  "executeMutation": Js.Json.t => Js.Promise.t('a),
+  "executeMutation":
+    Js.Json.t => Js.Promise.t(UrqlClient.UrqlExchanges.operationResult),
```
parkerziegler: These bindings better match `urql`'s API: https://github.com/FormidableLabs/urql/blob/774e09db8613d79e3e5cc6fca3d7f91526765f8e/src/components/Mutation.tsx#L21-L23.

src/components/UrqlSubscription.re
```diff
@@ -0,0 +1,64 @@
...
+    | (true, _, _) => Fetching
+    | (false, Some(data), _) => Data(data)
+    | (false, _, Some(error)) => Error(error)
+    | (false, None, None) => NotFound
+    };
+
+  {fetching: result##fetching, data, error, response};
+};
+
+type handler('a, 'b) =
```
parkerziegler: I did decide to make these named arguments. Also, the `urql` type here: https://github.com/FormidableLabs/urql/blob/master/src/components/Subscription.tsx#L13 specified the type of previous subscriptions as `any | void` (which I think may actually better be written as `any | undefined`). For this reason, we'll wrap this in an option to force users to handle the previous case where `~prevSubscriptions` is `undefined` (which is equivalent to Reason's `None`).

PR #50, opened 2019-04-29
Add npm installation instructions

oddlyfunctional:
After having some trouble installing with npm because of urql's version, I added the appropriate instructions to the README 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,13 @@ yarn add reason-urql urql
 yarn add graphql_ppx --dev
 ```
 
+If using npm, please use the following commands instead.
+
+```sh
+npm install -s reason-urql urql@0.2.1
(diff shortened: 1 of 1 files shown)


parkerziegler:
Thanks @oddlyfunctional! Yeah, we're in the middle of finishing off a `v1` which makes this process a lot easier. Previously, we included `urql` as a `peerDependency`, but now that `urql` has been bumped to a `v1` these bindings don't work with latest. The `v1` release we are planning (should be out in ~2 weeks), will fix this by including `urql` as a direct dependency. Still great to have this in for others using older versions!

parkerziegler:
@allcontributors[bot] add @oddlyfunctional for doc.

oddlyfunctional:
Thanks! 😄

Issue #34, opened 2019-03-25
Add examples for 'Query' component.


kiraarghy:
Like we have for the 'Mutation' component create an example of how this component can be used with reason-urql V1.

parkerziegler:
Closed by #56.

PR #54, opened 2019-05-07
#34 Query Example

kiraarghy:
Apologies for taking so long with this! 

This provides an example for our Query component. 

Happy to take feedback on the examples/code/comments!


examples/2-query/src/GetAll.re
```diff
@@ -0,0 +1,89 @@
+/* Opens our styles module */
+open GetAllStyles;
+
+type state = {
+  listOfPokemons: Js.Array.t(string),
```
parkerziegler: Can you talk about your decision to use `Js.Array.t`? I'm not super familiar with it, and IIRC Reason's native `array` will compile to a JS array directly, so you're safe binding to it in that way. A positive I can think of is that you wouldn't need to convert the array to a list to do the filtering below (since the native `Array` module doesn't have a `filter` operation). Happy to keep this as is, mostly just curious about your thinking!

examples/2-query/src/GetAll.re
```diff
@@ -0,0 +1,89 @@
...
+};
+
+type action =
+  | ChangeInput(string)
+  | SelectPokemon(string);
+
+/* Filters our pokemon list by our input */
+let filteredPokemonList = (~pokemonList: Js.Array.t(string), ~input: string) =>
+  Js.Array.filter(
+    pokemon => Js.String.includes(input, Js.String.toLowerCase(pokemon)),
```
parkerziegler: Do you need to call `.toLowerCase` on `input` as well to ensure case insensitivity?

examples/2-query/src/GetAll.re
```diff
@@ -0,0 +1,89 @@
+/* Opens our styles module */
+open GetAllStyles;
+
+type state = {
+  listOfPokemons: Js.Array.t(string),
+  textInput: string,
+  filteredList: Js.Array.t(string),
+  selectedPokemon: Js.Nullable.t(string),
```
parkerziegler: How would you feel about making this an `option(string)` type?

examples/2-query/src/GetAll.re
```diff
@@ -0,0 +1,89 @@
...
+          className=Styles.searchBox
+          value={self.state.textInput}
+          onChange={
+            event =>
+              self.send(ChangeInput(event->ReactEvent.Form.target##value))
+          }
+        />
+        <ul>
+          {
+            ReasonReact.array(
```
parkerziegler: In general I'd prefer the pipe operator here: `pokemonElementsArray(~pokemonList=self.state.filteredList) |> ReasonReact.array`.

examples/2-query/src/GetAllStyles.re
```diff
@@ -0,0 +1,38 @@
...
+      display(grid),
+      gridTemplateColumns([`fr(0.3), `fr(1.)]),
+      height(pct(100.)),
+      width(pct(100.)),
+      fontFamily("Roboto"),
+    ]);
+
+  let search =
+    style([
+      display(`flex),
```
parkerziegler: `bs-css` provides a convenience for this polymorphic variant (it's `flexBox`).

examples/2-query/src/Monster.re
```diff
@@ -0,0 +1,110 @@
...
+];
+let component = ReasonReact.statelessComponent("Example");
+
+let make = (~pokemon: string, _children) => {
+  ...component,
+  render: _self => {
+    /* We set up the query here as we need access to the pokemon
+       value passed in from GetAll */
+    let query = GetPokemon.make(~name=pokemon, ())##query;
+    let variables = Js.Dict.empty();
```
parkerziegler: So rather than needing to encode the variables yourself, `graphql_ppx` encodes them for you and allows you to access them as `##variables`. So you can do something like the following:

```reason
let request = GetPokemon.make(~name=pokemon, ());
let query = request##query;
let variables = request##variables; /* this will be a Js.Json.t */
```

examples/2-query/src/index.re
```diff
@@ -0,0 +1,77 @@
...
+|}
+];
+
+/* How many pokemon we want to get, in this case 151 */
+let numberOfPokemon = 151;
+
+/* Creates our query we pass into Query component */
+let query = GetAllPokemons.make(~first=numberOfPokemon, ())##query;
+
+let variables = Js.Dict.empty();
```
parkerziegler: Same comment here as above, just access `##variables`.

parkerziegler:
Also, from here on out would you be ok with opening PRs from branches off of the main Formidable repo (since you should have write access now). Hard to verify folks' branches w/ out doing a full clone, and it's nice to just `git fetch` to get the changes 🤗

examples/2-query/src/GetAll.re
kiraarghy: No decision behind this, I was just scrambling at stuff 🤷‍♀️ I'd rather rely on as much 'native' reason bindings as possible so have gone back to Array. Sucks having to convert to List but eh.

examples/2-query/src/GetAll.re
kiraarghy: yep good point!

examples/2-query/src/GetAll.re
kiraarghy: Done, much nicer!

examples/2-query/src/Monster.re
kiraarghy: Ah this explains a question I had about having to redefine variables a couple of times!

PR #55, opened 2019-05-08
Initial work on docs for v1.

parkerziegler:
This PR begins to add docs in preparation for a `v1` release. These don't need to be perfectly fleshed out before we tag a `v1`, but should at least document the core API. Ideally we'll have a Getting Started guide as well just to walk someone through the very basics of creating a client, wrapping their app in a `Provider`, and creating a `Query` component. Moving forward we can create more specific guides for things like `Mutation`s and `Subscription`s.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,58 +1,49 @@
 # reason-urql
 
 ![npm](https://img.shields.io/npm/v/reason-urql.svg)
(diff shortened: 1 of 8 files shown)


kiraarghy:
I really like this, seems really straightforwards!

PR #58, opened 2019-06-02
Migrate Provider and Mutation components to latest ReasonReact.

parkerziegler:
This PR begins the migration process to latest ReasonReact and the JSX 3 syntax. The majority of files in this diff are some stylistic and architectural changes to the example project for `Mutation`, which now shows uses of both the `Mutation` component and the `useMutation` hook.

**API Proposal**
The idea for the API is that `useMutation` will live alongside `useQuery` and `useSubscription` in a top-level `Hooks` module. It returns a two dimensional tuple, composed of the result of executing the mutation and the `executeMutation` function, which we get for free because of the way BuckleScript handles JS arrays of fixed length: https://bucklescript.github.io/docs/en/common-data-types#tuple.

```reason
open ReasonUrql;

let query = {|
    mutation likeDog($key: ID!) {
      likeDog(key: $key) {
        likes
      }
    }
  |};

[@react.component]
let make = (~id: string) => {
   let (res, executeMutation) = Hooks.useMutation(~query);

   /* Now you can use executeMutation in your UI. */
}
```

This keeps the API closely aligned to both `urql` and ReasonReact's hooks. cc/ @jaredly including you here in case you have any thoughts on the API.

### DIFF
--- a/bsconfig.json
+++ b/bsconfig.json
@@ -17,7 +17,7 @@
   ],
   "package-specs": [
     {
-      "module": "es6",
+      "module": "commonjs",
       "in-source": true
     }
   ],
--- a/bsconfig.json
+++ b/bsconfig.json
@@ -1,8 +1,9 @@
 {
   "name": "reason-urql",
   "reason": {
-    "react-jsx": 2
+    "react-jsx": 3
   },
+  "bsc-flags": ["-bs-super-errors"],
   "sources": [
     {
       "dir": "src",
@@ -16,7 +17,7 @@
   ],
   "package-specs": [
     {
-      "module": "commonjs",
+      "module": "es6",
       "in-source": true
     }
   ],
--- a/src/ReasonUrql.re
+++ b/src/ReasonUrql.re
@@ -15,3 +15,7 @@ module Request = UrqlRequest;
 module Error = UrqlCombinedError;
 
 module Exchanges = UrqlClient.UrqlExchanges;
+
+module Hooks = {
+  include UrqlUseMutation;
+};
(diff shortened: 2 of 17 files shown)


bsconfig.json
```diff
@@ -16,7 +17,7 @@
   ],
   "package-specs": [
     {
-      "module": "commonjs",
+      "module": "es6",
```
parkerziegler: This appears to be the reason for the build failure -> Jest doesn't like ES6 `import` syntax (a tale as old as time).

src/ReasonUrql.re
```diff
@@ -14,4 +14,8 @@ module Request = UrqlRequest;
 
 module Error = UrqlCombinedError;
 
-module Exchanges = UrqlClient.UrqlExchanges;
\ No newline at end of file
+module Exchanges = UrqlClient.UrqlExchanges;
+
+module Hooks = {
```
parkerziegler: Here's where we'll bring in future hooks.

src/UrqlProvider.re
```diff
@@ -1,5 +1,4 @@
-[@bs.module "urql"] external provider: ReasonReact.reactClass = "Provider";
-let component = ReasonReact.statelessComponent("Provider");
-
-let make = (~value: UrqlClient.t, children) =>
-  ReasonReact.wrapJsForReason(~reactClass=provider, ~props={"value": value}, children);
\ No newline at end of file
+[@bs.module "urql"] [@react.component]
```
parkerziegler: New ReasonReact syntax for `Provider`!

src/components/UrqlMutation.re
```diff
@@ -43,12 +43,17 @@ let urqlDataToRecord = (result: mutationRenderPropsJs('a)) => {
 [@bs.deriving abstract]
 type jsProps = {query: string};
 
+module MutationJs = {
```
parkerziegler: New restructured bindings for `Mutation`.

PR #59, opened 2019-06-04
Bind Query to new syntax. Add useQuery hook.

gugahoa:
Hello, everyone, excited make my first PR here! :)

## useQuery rationale
First of all, I want to highlight some considerations I had to take in making the useQuery binding not as good as the Query component.

### Currying Drawbacks
To start it all, the main driving force for how I decided to go with that API was the Drawback No. 3 here: https://bucklescript.github.io/docs/en/function#drawback

Because of BuckleScript currying, I couldn't get too sophisticated with the API before it messed up with the context React Hooks needs to properly function.

It was not possible to use a type signature such as:
```reason
let useQuery = (~query: string, ~variable=?, ~requestPolicy=?, ~pause=?, ()) => {};

// Doesn't even compile, as you can't force uncurrying with labeled args
// let useQuery = (. ~query: string, ~variable=?, ~requestPolicy=?, ~pause=?, ()) => {};

// Even with only one argument in the function, it would curry
let useQuery = args: richUseQueryResponse('a) => {};

// Cant force uncurrying here because of https://bucklescript.github.io/docs/en/function#pitfall
let useQuery = (.args): richUseQueryResponse('a) => {};
```

So in the end, the best option was to expose the binding directly, and use [@bs.deriving abstract] for the args record to maintain type safety in ReasonML land and not deal with Js object directly.

By the end of it, we have the bindings in this PR, and use it like this:
```reason
let args = useQueryArgs(~query, ~variables, ());
let (result, _executeQuery) = useQuery(args);
```

## Query component new syntax
This one was tricky not because of the complexity, or drawbacks from the language, it was because of a known issue with npm-link (https://github.com/facebook/react/issues/13991) that made testing it really hard. To solve it, you have to add to `webpack.config.js` resolve the following alias:
```javascript
  resolve: {
   alias: {
    react: path.resolve('./node_modules/react'),
     urql: path.resolve('./node_modules/urql'),
    "react-dom": path.resolve('./node_modules/react-dom'),
   }
  }
```

Maybe it would be interesting to add it to all examples already, so future users don't face the same issue?

I saw that in `3-mutation`, there's a resolve config too, like this:
```javascript
  resolve: {
    modules: [path.resolve(__dirname, "../../node_modules"), "node_modules"]
  },
```

This one seems more catch-all, and could help mitigate any issue that may arise with future libraries!


Schmavery:
Just an idea re: the useQuery currying -- what were your thoughts about redefining the useQuery function in reason so that it uses the useQuery external but provides a nicer interface? like this:

```reason
[@bs.module "urql"]
external useQuery: (useQueryArgs('a)) => useQueryResponse('b) = "useQuery";

let useQuery = (~query: string, ~variables=?, requestPolicy=?, pause=?, ()) =>
  useQuery(useQueryArgs(~query, ~variables?, ~requestPolicy?, pause?, ()));
```

Then people could call useQuery without needing to know about useQueryArgs

gugahoa:
I tried that approach, the currying will mess up the React Hook context that way too. It seems that uncurrying is a pre requisite for any React Hook in Reason.

gugahoa:
I just noticed that what I thought was the error (currying drawback) in the useQuery case, is not really the culprit. It was the webpack issue I mentioned that was giving me error and I thought it was the currying drawback.

Actually what you suggested, @Schmavery, is quite possible. I gave it a test again and it worked just fine. I'll commit later today the changes to make useQuery more consistent with useMutation implementation

examples/2-query/bsconfig.json
```diff
@@ -1,8 +1,9 @@
 {
   "name": "2-query",
   "reason": {
-    "react-jsx": 2
+    "react-jsx": 3
   },
+  "bsc-flags": ["-bs-super-errors"],
```
parkerziegler: Awesome, thanks for adding this! I forgot to track back after getting the initial `bsconfig`s setup, but this is wonderful to have 🚀

examples/2-query/src/GetAll.re
```diff
@@ -24,65 +24,63 @@ let filteredPokemonList = (~pokemonList: array(Js.String.t), ~input: string) =>
...
+  );
+
+  let pokemonElementsArray = (~pokemonList: array(string)) =>
+    Array.map(
+      pokemon =>
+        <li key=pokemon className=Styles.list>
+          <button
+            className=Styles.listButton
+            onClick={_event => dispatch(SelectPokemon(pokemon))}>
+            pokemon->ReasonReact.string
```
parkerziegler: Can we convert this to `React.string`?

examples/2-query/src/GetAll.re
```diff
@@ -24,65 +24,63 @@ let filteredPokemonList = (~pokemonList: array(Js.String.t), ~input: string) =>
...
+        className=Styles.searchBox
+        value={state.textInput}
+        onChange={
+          event =>
+            dispatch(ChangeInput(event->ReactEvent.Form.target##value))
         }
+      />
+      <ul>
+        {pokemonElementsArray(~pokemonList=state.filteredList)
+        ->ReasonReact.array}
```
parkerziegler: Can we convert this to `React.array`?

examples/2-query/src/GetAll.re
```diff
@@ -24,65 +24,63 @@ let filteredPokemonList = (~pokemonList: array(Js.String.t), ~input: string) =>
...
+      </ul>
+    </section>
+    {
+      switch (state.selectedPokemon) {
+      | Some(pokemon) => <Monster pokemon />
+      | None => ReasonReact.null
```
parkerziegler: One more 😅Can we change this to `React.null`?

examples/2-query/src/Monster.re
```diff
@@ -29,81 +30,78 @@ module GetPokemon = [%graphql
...
+[@react.component]
+let make = (~pokemon: string) => {
+  /* We set up the query here as we need access to the pokemon
+      value passed in from GetAll */
+  let request = GetPokemon.make(~name=pokemon, ());
+  let query = request##query;
+  let variables = request##variables;
+  let ({response}, _executeQuery) = useQuery(~query, ~variables, ());
+
+  <>
```
parkerziegler: Is this `Fragment` necessary? AFAICT pattern matching on `response` should only return a single parent node in the way it's currently written.

examples/2-query/src/Monster.re
```diff
@@ -29,81 +30,78 @@ module GetPokemon = [%graphql
...
+                  <h2 className=Styles.dexSubTitle>
+                    classification->ReasonReact.string
+                  </h2>
+                  {
+                    switch (height##maximum, weight##maximum) {
+                    | (Some(heightMax), Some(weightMax)) =>
+                      <div className=Styles.dexGrid>
+                        <p>
+                          ("Height: " ++ heightMax)
+                          ->ReasonReact.string
```
parkerziegler: +1 to converting all `ReasonReact.` methods in this JSX to `React.` methods.

examples/2-query/src/Monster.re
```diff
@@ -29,81 +30,78 @@ module GetPokemon = [%graphql
...
+                </div>
+              </div>
+            </section>
+          | (_, _, _, _, _) => ReasonReact.null
+          }
+        | None => ReasonReact.null
+        }
+      | Fetching => <div> "Loading"->ReasonReact.string </div>
+      | Error(error) => {
+        Js.log(error);
```
parkerziegler: Can we remove this `Js.log`?

examples/2-query/webpack.config.js
```diff
@@ -11,5 +11,12 @@ module.exports = {
   devServer: {
     open: true,
     contentBase: path.resolve(__dirname, "public")
+  },
+  resolve: {
+   alias: {
```
parkerziegler: How would you feel about opting for the module resolution we have setup in `3-mutation`, where we look in the project root before checking local `node_modules`?

src/components/UrqlQuery.re
```diff
@@ -1,70 +1,57 @@
...
+  external make:
+    (~query: string, ~variables: Js.Json.t, ~requestPolicy: string, ~children: queryRenderPropsJs('a) => React.element) =>
+    React.element =
+    "Query";
+}
+
 let urqlDataToRecord = (result: queryRenderPropsJs('a)) => {
-  let data = result##data |> Js.Nullable.toOption;
-  let error = result##error |> Js.Nullable.toOption;
+  let data = result->dataGet;
```
parkerziegler: Since these can still potentially be `undefined` / `null`, I _think_ we still need to call `Js.Nullable.toOption` on them (i.e. for `data` and `error`). So I imagine it looking like:
```suggestion
  let data = result->dataGet->Js.Nullable.toOption;
  let error = result->errorGet->Js.Nullable.toOption;
```

Does that make sense?

parkerziegler: With the `[@bs.optional]` annotation do these get wrapped in `option` automatically? Based on the pattern match below that appears to be what's happening, which is really nice!

src/components/UrqlQuery.re
```diff
@@ -1,70 +1,57 @@
...
-[@bs.deriving abstract]
-type jsProps = {
-  query: string,
-  [@bs.optional]
-  variables: Js.Json.t,
-  requestPolicy: string,
+[@react.component]
+let make = (
+  ~query: string,
+  ~variables: Js.Json.t,
```
parkerziegler: In `urql`, `variables` is an optional prop. I'd like to see it typed as optional both in the `make` signature here and in the `QueryJs` binding, i.e.

```suggestion
  ~variables: option(Js.Json.t)=?
```

in both places. Does that make sense or am I missing something?

src/components/UrqlQuery.re
```diff
@@ -1,70 +1,57 @@
...
+  ~requestPolicy: UrqlTypes.requestPolicy=`CacheFirst,
+  ~children: queryRenderProps('a) => React.element
```
parkerziegler: `pause` will also be available on `urql`'s `Query` component once the next version is released, would you mind adding it here and to the `QueryJs` binding as `~pause: option(bool)=?`.

src/hooks/UrqlUseQuery.re
```diff
@@ -0,0 +1,52 @@
...
+
+[@bs.deriving abstract]
+type useQueryArgs('a) = {
+  query: string,
+  [@bs.optional] variables: 'a,
+  [@bs.optional] requestPolicy: UrqlTypes.requestPolicy,
+  [@bs.optional] pause: bool
+}
+
+type partialOperationContextFunc = option(UrqlTypes.partialOperationContext) => unit;
```
parkerziegler: I'd say let's just define this type inline on L17 below (right in the tuple as the second argument) since it's a somewhat unique one-off.

src/hooks/UrqlUseQuery.re
```diff
@@ -0,0 +1,52 @@
...
+  {fetching, data, error, response};
+};
+
+[@bs.module "urql"]
+external useQueryJs: (useQueryArgs('a)) => useQueryResponseJs('b) = "useQuery";
+
+let useQuery = (~query, ~variables=?, ~requestPolicy=?, ~pause=?, ()) => {
+  let args = useQueryArgs(~query, ~variables=?variables, ~requestPolicy=?requestPolicy, ~pause=?pause, ());
+  let (state, executeQuery) = useQueryJs(args);
+  let state = state |> useQueryResponseToRecord;
```
parkerziegler: Would you mind creating a different name for this `let` binding rather than overriding the `state` binding from the previous line? It's totally fine practice, but I want to keep this as beginner friendly for JS devs as possible and I could see this being a stumbling block.

src/hooks/UrqlUseQuery.rei
```diff
@@ -0,0 +1,10 @@
+type partialOperationContextFunc = option(UrqlTypes.partialOperationContext) => unit;
```
parkerziegler: ❤️Thanks for generating a `.rei` file! I've always had trouble w/ this due to the CLI mucking up the `[@bs]` attributes. Did you hand write this or did the CLI generate it?

examples/2-query/webpack.config.js
gugahoa: I don't know much about Webpack configuration, when I tried to configure the resolve in `2-query` the same way it was done in `3-mutation`, it gave me some errors.
But I believe that it would be good to do it like in `3-mutation`

src/components/UrqlQuery.re
gugahoa: It does! And also play nice with JavaScript object interop. So we have the benefits of type safety and type checking, while being easy to interop with JS obj.
https://bucklescript.github.io/docs/en/object#record-mode

src/hooks/UrqlUseQuery.rei
gugahoa: Wrote it by hand :)

gugahoa:
Thanks for the review @parkerziegler!
I changed everything you suggested, except for the webpack config as I don't know much about it, and it gave me some errors when I tried to copy-paste from `3-mutation`

parkerziegler:
@all-contributors add @gugahoa for code and example.

parkerziegler:
@all-contributors add @Schmavery for ideas.

parkerziegler:
@all-contributors add @kiraarghy for example.

PR #60, opened 2019-06-05
UrqlUseSubscription

kiraarghy:
1. Created UrqlUseSubscription.


src/components/UrqlSubscription.re
```diff
@@ -15,50 +14,61 @@ type subscriptionRenderProps('a) = {
...
-  );
-};
\ No newline at end of file
+      ~query: string,
+      ~variables: option(Js.Json.t),
+      ~handler: option(handler('a, 'b)),
+      ~children: subscriptionRenderProps('a) => React.element,
+    ) =>
+  switch (variables, handler) {
+  | (Some(v), Some(h)) =>
```
Schmavery: (I think I'm in direct violation of the title by commenting but in the off-chance it saves you some time, I think you can pass through an optional named argument in jsx with `<SubscriptionJs variables=?v />`)

parkerziegler:
@kiraarghy I pulled locally and I _think_ the issue had to do with `webpack` getting all mixed up with which copies of which dependencies to use. Specifically, we're in a weird spot in our repo setup because we `yarn link` to the `src` directory to pull `reason-urql` in. However, this means our examples get _all of the linked dependencies' dependencies_ 😞(i.e. from local `node_modules` and from `node_modules/reason-urql/node_modules/*` brought in by `yarn link`). To resolve this, we use `webpack`'s module resolution to tell it where to look for dependencies. That way, we guarantee that it only ever uses a single copy of a dependency, like `urql` or `react`, at runtime.

The commit I pushed up also moves us more down the path of `[@bs.deriving abstract]`. Overall things are looking ace and this is pretty darn close. I'm itching to get it merged ASAP so we can release a v1 before I go on PTO on Wednesday 😱But we still have to circle back for at least cursory docs so that may be a little too soon for release.

kiraarghy:
Okay I think this is ready for review now!

src/components/UrqlSubscription.re
kiraarghy: Thank you so much, so much simpler! 😄

examples/5-subscription/src/app/Messages.re
```diff
@@ -53,31 +53,28 @@ module Styles = {
...
-        | Error(_e) => <div> "Error"->str </div>
-        | NotFound => <div> "Not Found"->str </div>
-        }
-      }
-    </Subscription>,
-};
\ No newline at end of file
+[@react.component]
+let make = () =>
+  <Subscription query variables=None handler=None>
```
gugahoa: This is a breaking change, variables and handler were optional args, now they're required and with type optional
Read more at `UrqlSubscription.re` about this

src/components/UrqlSubscription.re
```diff
@@ -15,50 +14,46 @@ type subscriptionRenderProps('a) = {
...
 
+type handler('a, 'b) =
+  (~prevSubscriptions: option('a), ~subscription: 'b) => 'a;
+
+module SubscriptionJs = {
+  [@bs.module "urql"] [@react.component]
+  external make:
+    (
+      ~query: string,
+      ~variables: option(Js.Json.t),
```
gugahoa: When you declare a named argument with type `option('a)`, it does not mean the argument is optional to the caller, it means it receives a `option('a)` type.
I believe what you wanted was: `~variables: Js.Json.t=?`
```suggestion
      ~variables: Js.Json.t=?,
```

src/components/UrqlSubscription.re
```diff
@@ -15,50 +14,46 @@ type subscriptionRenderProps('a) = {
...
+      ~handler: option(handler('a, 'b)),
```
gugahoa: The same as above:
```suggestion
      ~handler: handler('a, 'b)=?,
```

src/components/UrqlSubscription.re
```diff
@@ -15,50 +14,46 @@ type subscriptionRenderProps('a) = {
...
-    };
-
-  ReasonReact.wrapJsForReason(
-    ~reactClass=subscriptionComponent, ~props, result =>
-    result |> urqlDataToRecord |> children
-  );
-};
\ No newline at end of file
+      ~query: string,
+      ~variables: option(Js.Json.t),
```
gugahoa: Same as `SubscriptionJs`
```suggestion
      ~variables: Js.Json.t=?,
```

src/components/UrqlSubscription.re
```diff
@@ -15,50 +14,46 @@ type subscriptionRenderProps('a) = {
...
+      ~handler: option(handler('a, 'b)),
```
gugahoa:
```suggestion
      ~handler: handler('a, 'b)=?,
```

src/components/UrqlSubscription.re
```diff
@@ -15,50 +14,46 @@ type subscriptionRenderProps('a) = {
...
+      ~children: subscriptionRenderProps('a) => React.element,
+    ) =>
+  <SubscriptionJs query ?variables ?handler>
```
gugahoa:
```suggestion
  <SubscriptionJs query variables handler>
```

kiraarghy:
Hey I'm probably not going to be able to have a look at this till tomorrow evening, given @parkerziegler's desire to get the next release I'm happy for someone else to finish it off?

parkerziegler:
@kiraarghy I won't have time for release since I'm full time on client work and can only work on this stuff in the evening. I'll try to carry it through sometime tomorrow morning so we can have a v1 release ready to go for next week. We may make a second beta release as well to test these things before cutting a v1.

src/components/UrqlSubscription.re
parkerziegler: Agreed with @gugahoa here. Going to commit this suggestion.

src/components/UrqlSubscription.re
parkerziegler: +1 here, thanks @gugahoa.

src/components/UrqlSubscription.re
parkerziegler: @gugahoa with your changes in place, i.e. `variables` as `~variables: Js.Json.t=?` and `~handler: handler('a, 'b)=?` I believe this can stay as is now cc/ @Schmavery.

examples/5-subscription/src/app/Messages.re
gugahoa: After the changes, this should work as before:
```suggestion
  <Subscription query>
```

src/UrqlTypes.re
```diff
@@ -82,4 +82,21 @@ type operationResult = {
   data: operationData,
   [@bs.optional]
   error: UrqlCombinedError.t,
-};
\ No newline at end of file
+};
+
+/* The signature of the structure created by calling `.make()` on a `graphql_ppx` module. */
+type request('response) = {
```
parkerziegler: We'll want to use this type in `useQuery` and `useMutation` cc/ @gugahoa @Schmavery.

src/components/UrqlSubscription.re
```diff
@@ -1,11 +1,10 @@
-[@bs.module "urql"]
-external subscriptionComponent: ReasonReact.reactClass = "Subscription";
-
+[@bs.deriving abstract]
```
parkerziegler: There are some follow up changes to make here, but I'm going to make a separate issue to track making the components type safe in the same manner as the hooks.

src/hooks/UrqlUseSubscriptionWithHandler.re
```diff
@@ -0,0 +1,72 @@
...
+let useSubscriptionResponseToRecordWithHandler =
+    (result: useSubscriptionWithHandlerResponseJs('acc)) => {
+  let data = result->dataGet;
+  let error = result->errorGet;
+  let fetching = result->fetchingGet;
+
+  let response =
+    switch (fetching, data, error) {
+    | (true, None, _) => UrqlTypes.Fetching
+    | (true, Some(d), _) => Data(d)
```
parkerziegler: So `urql` has a strange case where `fetching` is never set to `false` on subscriptions (I've pinged folks internally to ensure that's intended behavior). To handle this, I needed to add this extra case to this pattern matching to check if we are receiving data even as `fetching` remains true.

src/hooks/UrqlUseSubscriptionWithHandler.re
```diff
@@ -0,0 +1,72 @@
...
+  {fetching, data, error, response};
+};
+
+let useSubscriptionWithHandler =
+    (
+      request: UrqlTypes.request('response),
+      handler: UrqlTypes.parsedHandler('acc, 'response),
+    ) => {
+  let parse = request##parse;
+  let parsedHandler =
```
parkerziegler: So the idea around `parsedHandler` is sort of interesting. Basically, the data that `urql` returns to you varies based on whether or not you provide a `handler` to `useSubscription`. If you **don't** provide a `handler`, `urql` will provide the raw response of executing the subscription (a `Js.Json.t` that we `parse` into `'response`). If you _do_ provide a `handler` `urql` will return whatever data structure you are using to accumulate your subscription data (referred to here as `'acc`). `parsedHandler`, realizing this, parses each `subscription` as it comes in, and then passes it on to the `handler` supplied by the user. This ensures 1) proper type inference of subscription responses as they come in, and 2) proper type inference for the `data` ('`acc') returned by `useSubscription`. You can see an example of this in the Subscription example.

parkerziegler:
@Schmavery @gugahoa @kiraarghy if any of you have some time to review the latest changes I'd really appreciate it 🙏. I took the guidance of @Schmavery's two PRs #68 and #69 to inform this one, and the type inference should be 💯 here. Let me know thoughts about the additional `useSubscriptionWithHandler` hook, which is mostly there to reduce polymorphism due to the way `urql` passes back different data based on whether you call `useSubscription` with a handler or not.

src/hooks/UrqlUseSubscriptionWithHandler.re
Schmavery: This is a cool concept, seems like some sort of reducer for subscriptions? Out of curiosity (maybe this is out of scope), how would someone have a subscription that needs to be initialized by a query?
Like if I have an `onlineUsers` query and some sort of `onlineUserUpdate` subscription that adds or removes users from that list.

parkerziegler: Yep, pretty much! `urql` allows you to pass a `handler` and will let you accumulate previous subscriptions with whatever data structure you want. The crux is that, when you pass a `handler`, `urql` returns the accumulator you use to maintain data, whereas when you don't, it just returns the raw subscription result, hence the need for two separate hooks here (to get the type inference) right.

In terms of a `subscription` initialized by a `query`, the only way I can think to do it is by having the component that calls `useQuery` higher up in the tree than the one that calls `useSubscription`, and passing the result of `useQuery` into the child that can then pass it into `useSubscription`. @kitten may have other thoughts, I think there's some discussion on this over in the `urql` issues.

kitten: We were thinking of the state being combined. The initial state of the subscription being the queries initial result.

We've purposefully haven't integrated the two because that can also be done with:

- a normalised cache
- a custom hook wrapping useQuery and use Subscription
- or; local component state

So we wanted to leave this choice up to the user, since it'd otherwise feel like we were over stepping the boundary of what should and can be generalised

gugahoa:
This PR got a little bit too big for me to effectively review, is there anywhere I can find you guys online to chat a little bit while reviewing to make sure I don't miss any context?

Meanwhile I'll comment what I gathered so far

src/UrqlTypes.re
```diff
@@ -82,4 +82,21 @@ type operationResult = {
...
+  "parse": Js.Json.t => 'response,
+  "query": string,
+  "variables": Js.Json.t,
+};
+
+/* The type of the handler function passed to Subscription and useSubscription.
+   `handler` corresponds to the type of the handler function _before_ the latest subscription has been parsed.
+   `parsedHandler` corresponds to the type of the handler function _after_ the latest subscription has been parsed. */
+type handler('acc) =
+  (~prevSubscriptions: 'acc, ~subscription: Js.Json.t) => 'acc;
```
gugahoa: Shouldn't this be `option('acc)`?
```suggestion
  (~prevSubscriptions: option('acc), ~subscription: Js.Json.t) => 'acc;
```

src/hooks/UrqlUseSubscription.re
```diff
@@ -0,0 +1,59 @@
...
+type useSubscriptionResponse('response) = {
+  fetching: bool,
+  data: option('response),
+  error: option(UrqlCombinedError.t),
+  response: UrqlTypes.response('response),
+};
+
+[@bs.module "urql"]
+external useSubscriptionJs:
+  useSubscriptionArgs => array(useSubscriptionResponseJs) =
```
gugahoa: I believe it's possible do to away with `useSubscriptionWithHandlerJs` if we change the signature here and below with something like: `(~handler: UrqlTypes.handler('acc)=?, useSubscriptionArgs) => array(useSubscriptionResponseJs)`

I'm working on an example of how it could work

src/hooks/UrqlUseSubscriptionWithHandler.re
```diff
@@ -0,0 +1,72 @@
...
+  query: string,
+  [@bs.optional]
+  variables: Js.Json.t,
+};
+
+[@bs.deriving abstract]
+type useSubscriptionWithHandlerResponseJs('acc) = {
+  fetching: bool,
+  [@bs.optional]
+  data: 'acc,
```
gugahoa: Possible issue with type safety

src/UrqlTypes.re
parkerziegler: Yep, good catch! The way I understand `~prevSubscriptions`, it will _always_ be `undefined` the first time it runs, so wrapping it in `option` is correct 👍 Just missed this one.

src/hooks/UrqlUseSubscription.re
parkerziegler: Ah gotcha, so we'd have one `useSubscriptionJs` that takes an optional argument for `~handler` for when `useSubscriptionWithHandler` binds to it. I could see that working!

src/hooks/UrqlUseSubscriptionWithHandler.re
parkerziegler: Can you explain a little more? I checked the type inference on the two hooks and they _appeared_ to be working as expected. `useSubscription` will infer the type of `data` as the result of calling `parse` on the `Js.Json.t` returned by `urql` (written as `'response`). `useSubscriptionWithHandler` will infer the type of the accumulator `'acc` you use in your `handler` function. We also already parse each `Js.Json.t` subscription coming back from the API before handing it off to `handler`, so the inference should work as:

1. Infer the type of `'response` from calling `parse`.
2. Infer the type of `'acc` via the `handler` function (i.e. does a user supply an array, a `Js.Dict`, etc. for accumulating new subscriptions).

This piece is somewhat complex tho and I could definitely be missing something. Thannks as always for a thorough review ❤️

parkerziegler:
@gugahoa agreed that the diff is getting pretty unreadable w/ all this commenting. If it works for you, I'll make any changes you'd like to see by 5pm PST today, we can merge this, and then we can tackle follow up questions in a new PR? We don't currently have any other channels setup online for discussion, tho I am considering getting a Spectrum set up for `urql` that could also encompass `reason-urql`.

src/hooks/UrqlUseSubscriptionWithHandler.re
gugahoa: You're completely right! I hadn't thought that `useSubscriptionWithHandlerJs` is safely wrapped through `useSubscriptionHandler`

Maybe it would be good to comment this for future contributors to take notice? Because it can still be used in a non-type safe manner, but not by the library users, only inside this module. I believe it would be good to refactor a little bit so that the unsafe surface area is minimized

gugahoa:
@parkerziegler it would be awesome to have a channel where we can discuss a little bit before consolidating the discussion in github comments.

As for changes, how about only merging `useSubscription` and `Subscription`, and move `useSubscriptionWithHandler` to a new PR. What do you think?

parkerziegler:
@gugahoa I'm a little reluctant to split them just because the example uses `useSubscriptionWithHandler` and it'd be a pain to break it out now. I think I'd prefer to merge both and then open a separate issue where we can chat additional next steps for it. We still need to bring in these type safety checks for the components as well before cutting a `v1`, so I think we'll have some opportunity to follow up. Is that ok?

gugahoa:
Sure! It's ok for me

Schmavery:
@parkerziegler I just had a last-minute idea re: handler. (I agree that this PR is getting hairy so feel free to ignore it for now).

Assuming a signature of useSubscription that looks something like this:
```reason
let useSubscription: (
  ~handler: UrqlTypes.parsedHandler('acc, 'response)=?,
  ~request: UrqlTypes.request('response)) =>
  useSubscriptionResponse('acc);
```

What if we gave subscriptions a default handler of `(_, x) => x`?  Like an identity function that preserves the type of the subscription. Then the return type would be the same as 'response when you don't provide a handler, and otherwise it would be the return type of handler.

Not 100% sure it will work but it could let us avoid having 2 entrypoints for subscriptions?

gugahoa:
@schmavery seems like a great idea. I was trying to achieve polymorphic return with GADT, but this approach is a lot simpler and I can see it working.

gugahoa:
From what I gathered, it doesn't seem possible with a helper function:
```reson
type request('b) =
  | Box('b);

let barHelper = (~handler: (option('a), 'b) => 'a, ~request: request('b)) => {
  let Box(r) = request;
  handler(None, r)
}

let bar = (~handler: option((option('a), 'b) => 'a)=?, ~request: request('b), ()) => {
  switch (handler) {
  | None => barHelper(~handler=(_, x) => x, ~request)
  | Some(handler_fn) => barHelper(~handler=handler_fn, ~request)
  }
}```

The inferred type of `bar` is:
```reason
let bar: (~handler: (option('b), 'b) => 'b=?, ~request: request('b), unit) => 'b = <fun>;
```

Schmavery:
Hmm, interesting, thanks for trying this this. I'm gonna see if I can get something that works but feel free to ignore this for now guys

gugahoa:
@Schmavery @parkerziegler here an example on how we can achieve polymorphic returns for `useSubscription` https://reasonml.github.io/en/try?rrjsx=true&reason=C4TwDgpgBAThCOBXCBnYAKA5AIwJRQF4AoKKAHygCEB7ADyzwG4ijRIoALAQwDsATADYQYWLgBooOCZgDG+YqQoAJXoIjp01MMACW1HqNzS8hAHySuuAFydVQkVt37Dx6ZZLkoAOWor+Qm25-YXQAfVdJJhY2aEQUCABlRGwUGRgdbT0DTHFI6TlCD3QAPyC1GEC7EJyI2SMoYrgkVGAbJuQ0Blx5c1lmIiFgKAAzampC0nQYqFzp7AlpmQlSqorbYJFc+ag5ZfaWtoQOjDxrHbMoAG8PUkGqOnQYeVgjluZSUhQAdx1gGQ4oOgyvZ8NcPp4-GogatQsMeD11uVYQYfDwIBInjdPD5IUILjAsQBfd5QYkeIijaglYHCAg4qp7V5oAg0egAVm6zEp1NWdN8DIa+2ZrPQACI2aLORSxjyNgRcep0FwZEsoHwuMBLBcwVBvr9-oDlQUdRRUdACOYANrqzUAXQ8FAS1AAtuoBAiBFAAAJQa0arj20iEoiE+qNJnAFkPDm4RhAA

parkerziegler:
@gugahoa @Schmavery I merged this since it was getting pretty darn big, and opened a new issue [here](https://github.com/FormidableLabs/reason-urql/issues/72) for us to discuss API options. Cheers!

Issue #67, opened 2019-06-13
Hooks type safety idea


Schmavery:
I'm pretty excited to try to use this in some projects, but just realized after reading through #64 how the typing works for the hooks api. @gugahoa brought up some really good points (I thought) about how this could potentially be improved.
I think it would be hard for me to use this in real projects if I couldn't rely on the type of the response (though @parkerziegler did provide a clever workaround in that thread!).

I was thinking about this a bit wondering what you thought about a slightly different api.
Taking `useQuery` as an example, it seems from one of the examples that this is a pattern for using it:
https://github.com/FormidableLabs/reason-urql/blob/14bb74d1d003b9c9a058b64f65f4b11312f3dfa7/examples/2-query/src/Monster.re#L39-L41
Where we use graphql_ppx to create a query and variables object and then pass them into useQuery. I was wondering if there would be any downsides to simply passing in the entire result from calling `GetPokemon.make` (in this example)?

To be clear, the usage would look something like this:
```reason
  let request = GetPokemon.make(~name=pokemon, ());
  let ({response}, _executeQuery) = useQuery(request, ());
```

It seems like if you did that, the type of useQuery could be:
```reason
let useQuery:
  ({
    .
    "parse": Js.Json.t => 'response,
    "query": string,
    "variables": 'vars,
  }, 
  ~requestPolicy: UrqlTypes.requestPolicy=?,
  ~pause: bool=?,
  unit) =>
  useQueryResponse('response);

```

instead of
https://github.com/FormidableLabs/reason-urql/blob/14bb74d1d003b9c9a058b64f65f4b11312f3dfa7/src/hooks/UrqlUseQuery.rei#L10

The takeaway, as far as the types are concerned, would be that the type checker would be able to know what the return type of useQuery is.

Implementation-wise, inside of useQuery you could access the query and variables parameters as before, with the added bonus that you could access the parse function too, if you wanted to do any runtime validation/translation of the returned object (which would have the added benefit of supporting graphql_ppx features like the super-handy [@bsDecoders](https://github.com/mhallin/graphql_ppx#custom-field-decoders)).

I haven't had a chance to check if the same technique could work for mutations/subscriptions yet but wanted to get people's thoughts in case this would be a big enough signature change to warrant trying to squeeze in before v1?

parkerziegler:
[...]

PR #68, opened 2019-06-16
Typed useQuery MVP

Schmavery:
Implementation of https://github.com/FormidableLabs/reason-urql/issues/67 for useQuery.
It seems to work out pretty cleanly for useQuery, my followup PR to useMutation is slightly worse.
Example 2-query builds and seems to work fine after the change.

(It seems like I have a different version of refmt from the normal maintainers of the project, so I did my best to keep the reformatting changes to unrelated code out of the diff to keep it small/readable)

Thoughts?

(Edit: I was thinking, it's probably worth taking a look at the component bindings to see if they could benefit from a similar change, as the use of the graphql parse function here probably introduces some difference in parsing behaviour if it's not being used on the components. I haven't done it yet because I'm mostly interested in using the hooks.)

### DIFF
--- a/examples/2-query/package.json
+++ b/examples/2-query/package.json
@@ -6,6 +6,7 @@
   "author": "Kara Stubbs <kara.stubbs@formidable.com>",
   "license": "MIT",
   "scripts": {
+    "build": "bsb -make-world",
     "start": "bsb -make-world -w",
     "clean": "bsb -clean-world",
     "start:demo": "webpack-dev-server --hot"
--- a/src/hooks/UrqlUseQuery.re
+++ b/src/hooks/UrqlUseQuery.re
@@ -24,12 +24,17 @@ type useQueryState('a) = {
 }
 type useQueryResponse('a) = (useQueryState('a), partialOperationContextFn);
 
-let useQueryResponseToRecord = (result: useQueryStateJs('a)) => {
-  let data = result->dataGet;
+let useQueryResponseToRecord =
+    (parse: Js.Json.t => 'response, result: useQueryStateJs('a)) => {
+  let data =
+    switch (result->dataGet) {
+    | None => None
+    | Some(d) => Some(parse(d))
+    };
   let error = result->errorGet;
   let fetching = result->fetchingGet;
 
-  let response: UrqlTypes.response('a) =
+  let response: UrqlTypes.response('response) =
     switch (fetching, data, error) {
     | (true, _, _) => Fetching
     | (false, Some(data), _) => Data(data)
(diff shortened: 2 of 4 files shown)


examples/2-query/package.json
```diff
@@ -6,6 +6,7 @@
   "author": "Kara Stubbs <kara.stubbs@formidable.com>",
   "license": "MIT",
   "scripts": {
+    "build": "bsb -make-world",
```
Schmavery: I'm happy to remove this if you'd like -- I kept accidentally trying to run `yarn build` and saw that the root project has a "build" script like this so thought it might be ok.

parkerziegler: No I think this is great! Sometimes I want to build one-off to check the compiler, other times I like to pass `-w` when hacking away quickly on the examples.

src/hooks/UrqlUseQuery.rei
```diff
@@ -7,4 +7,16 @@ type useQueryState('a) = {
...
 type useQueryResponse('a) = (useQueryState('a), partialOperationContextFn);
 
-let useQuery: (~query: string, ~variables: 'a=?, ~requestPolicy: UrqlTypes.requestPolicy=?, ~pause: bool=?, unit) => useQueryResponse('b)
+let useQuery:
+  (
+    ~request: {
+                .
+                "parse": Js.Json.t => 'response,
+                "query": string,
+                "variables": 'vars,
```
parkerziegler: AFAIK the type of `variables` returned by calling make is always a `Js.Json.t`: https://github.com/mhallin/graphql_ppx#send-queries. Any reason to make it a type arg here rather than just specifying `Js.Json.t`?

src/hooks/UrqlUseQuery.re
```diff
@@ -24,12 +24,17 @@ type useQueryState('a) = {
 }
 type useQueryResponse('a) = (useQueryState('a), partialOperationContextFn);
 
-let useQueryResponseToRecord = (result: useQueryStateJs('a)) => {
-  let data = result->dataGet;
+let useQueryResponseToRecord =
+    (parse: Js.Json.t => 'response, result: useQueryStateJs('a)) => {
```
parkerziegler: I think the idea with `useQueryStateJs` is that the type argument here `'a` corresponds to the raw response from the GraphQL request _before_ it's been parsed by `graphql_ppx`. I've been using `'a`, `'b`, etc. with the thought that it's more OCaml-y, but I like what you did with `'response`, so perhaps I'll make these type signatures more descriptive in a follow up PR. This looks great though!

parkerziegler:
@Schmavery I like this a lot! Thanks for the help 🤗 See my one question about `Js.Json.t` vs. using `'vars`. Once we get that cleared this looks ready to merge!

Schmavery:
@parkerziegler not 100% sure what you were saying about useQueryStateJs, I think it's still corresponding to the type before the graphql request, but I fixed the variable type :)

src/hooks/UrqlUseQuery.re
```diff
@@ -24,12 +26,17 @@ type useQueryState('a) = {
...
+  let data =
+    switch (result->dataGet) {
+    | None => None
+    | Some(d) => Some(parse(d))
```
gugahoa: I believe that it's now safe to remove the type parameter from `useQueryStateJs` and `useQueryResponseJs`, as the only type `option('a)` that satisfies this switch is `option(Js.Json.t)`, as the only type accepted by `parse` is `Js.Json.t`

This switch could also become: `result->dataGet->Belt.Option.map(parse)`, but I don't know about the thoughts of using too much piping in this project, as it can be a little alien to newcomers.

parkerziegler: Good point @gugahoa that type parameter is unnecessary now. @Schmavery do you mind implementing this before we merge? I'm cool with piping and `Belt.Option.map`, I think it feels a little nicer than this `switch` if a little intimidating at first for newcomers.

src/hooks/UrqlUseQuery.re
```diff
@@ -8,10 +8,12 @@ type useQueryStateJs('a) = {
 [@bs.deriving abstract]
 type useQueryArgs('a) = {
```
parkerziegler: Can also remove the type parameter here now!

Schmavery:
Thanks for the review, I removed those parameters!

parkerziegler:
No, thank you! This is a great API change. I'll go and take a look at `useMutation` now.

parkerziegler:
@all-contrbituors add @Schmavery for code, example.

PR #69, opened 2019-06-16
Typed useMutation

Schmavery:
The useMutation-counterpart of https://github.com/FormidableLabs/reason-urql/pull/68.

I wanted to see if the easy change for useQuery could work for useMutation as well. For the most part, it was straightforward, but I had to make a second little change to the api because the Module.make() function (the one generated by graphql_ppx that returns the {"query", "variables", "parse"} object) requires the variables to be passed in when you call it rather than as a payload to the executeMutation call.

The downside of this is that it changes the useMutation interface between the component and the hook and it's not the same interface as the js urql. One workaround we could do is passing the makeWithVariables function itself to useMutation so that usage would look like this:
```reason
let (_, executeMutation) = Hooks.useMutation(~make=LikeDog.makeWithVariables);
executeMutation({"key": id});
```

I wasn't sure which design was better so just I went with one of them in the PR.

I noticed that the queries in 3-mutation weren't using graphql_ppx so I converted one of them to use it, and left another as an example of how this api could be used if you didn't want a dependency on graphql_ppx. It's a little worse than before I guess, but not terrible, I thought.

### DIFF
--- a/examples/3-mutation/src/Dog.re
+++ b/examples/3-mutation/src/Dog.re
@@ -1,13 +1,15 @@
 open ReasonUrql;
 
 module Mutations = {
-  let likeDog = {|
+  module LikeDog = [%graphql
+    {|
     mutation likeDog($key: ID!) {
       likeDog(key: $key) {
         likes
       }
     }
-  |};
+  |}
+  ];
 
   let patDog = {|
     mutation patDog($key: ID!) {
@@ -56,22 +58,28 @@ let make =
       [|id|],
     );
 
-  let (_, executeMutation) = Hooks.useMutation(~query=Mutations.treatDog);
+  let (_, executeLikeMutation) =
+    Hooks.useMutation(~request=Mutations.LikeDog.make(~key=id, ()));
+
+  let (_, executeTreatMutation) =
+    Hooks.useMutation(
+      ~request={
+        "query": Mutations.treatDog,
+        "variables": payload,
+        "parse": x => x,
+      },
+    );
 
   <div className=DogStyles.container>
     <img src=imageUrl alt=name className=DogStyles.image />
     <h3 className=DogStyles.title> {j|$name|j}->React.string </h3>
     <div className=DogStyles.buttons>
-      <Mutation query=Mutations.likeDog>
-        ...{({executeMutation}) =>
-          <EmojiButton
-            emoji={j|👍|j}
-            count={string_of_int(likes)}
-            hex="48a9dc"
-            onClick={_ => executeMutation(Some(payload)) |> ignore}
-          />
-        }
-      </Mutation>
+      <EmojiButton
+        emoji={j|👍|j}
+        count={string_of_int(likes)}
+        hex="48a9dc"
+        onClick={_ => executeLikeMutation() |> ignore}
+      />
       <Mutation query=Mutations.patDog>
         ...{({executeMutation}) =>
           <EmojiButton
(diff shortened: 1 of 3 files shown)


Schmavery:
There are probably some bits of feedback from #68 that I should migrate to this PR when I get a chance, but I'd still be curious to get opinions on the general api

examples/3-mutation/src/Dog.re
```diff
@@ -56,22 +58,28 @@ let make =
       [|id|],
     );
 
-  let (_, executeMutation) = Hooks.useMutation(~query=Mutations.treatDog);
+  let (_, executeLikeMutation) =
+    Hooks.useMutation(~request=Mutations.LikeDog.make(~key=id, ()));
+
+  let (_, executeTreatMutation) =
+    Hooks.useMutation(
+      ~request={
```
parkerziegler: Is the idea here to show a way to use `useMutation` without `graphql_ppx`? If so, I like it!

src/hooks/UrqlUseMutation.re
```diff
@@ -22,12 +22,17 @@ external useMutationJs:
   (~query: string) => (useMutationResponseJs('a), executeMutation) =
   "useMutation";
 
-let useMutationResponseToRecord = (result: useMutationResponseJs('a)) => {
-  let data = result->dataGet;
+let useMutationResponseToRecord =
+    (parse: Js.Json.t => 'response, result: useMutationResponseJs('a)) => {
```
parkerziegler: Similar to how we did things on `useQuery`, I believe `useMutationResponseJs('a)` can lose the type parameter and we can just type `data` as `Js.Json.t`.

src/hooks/UrqlUseMutation.re
```diff
@@ -22,12 +22,17 @@ external useMutationJs:
...
+  let data =
```
parkerziegler: Also similar to `useQuery` let's opt for `Belt.Option.map` like @gugahoa suggested rather than this `switch`.

src/hooks/UrqlUseMutation.re
```diff
@@ -22,12 +22,17 @@ external useMutationJs:
...
+    | None => None
+    | Some(d) => Some(parse(d))
+    };
   let error = result->errorGet;
   let fetching = result->fetchingGet;
 
-  let response: UrqlTypes.response('a) =
+  let response: UrqlTypes.response('response) =
     switch (fetching, data, error) {
     | (true, _, _) => Fetching
```
parkerziegler: One more "similar to `useQuery`" item – I liked how you stripped specifying the type here and instead supplied `UrqlTypes.Fetching` on the first match of the pattern matching (L37) to locally open that variant type. Let's opt for that here as well 🙌

src/hooks/UrqlUseMutation.re
```diff
@@ -38,9 +43,18 @@ let useMutationResponseToRecord = (result: useMutationResponseJs('a)) => {
...
+    (
+      ~request: {
+         .
+         "query": string,
+         "variables": Js.Json.t,
+         "parse": Js.Json.t => 'response,
+       },
+    ) => {
+  let (useMutationResponseJs, executeMutation) =
+    useMutationJs(~query=request##query);
```
parkerziegler: Since we only have a single parameter for `useMutationJs`, do you mind switching that `external` to _not_ use a labelled argument for `query`? I know I had it that way originally 😅, but I don't see much use in it now.

Schmavery:
@parkerziegler I think that's everything

Issue #72, opened 2019-06-20
Subscription and useSubscription API discussion.


parkerziegler:
@Schmavery @gugahoa I'm opening this issue here as a place for us to track API ideas for `Subscription` and `useSubscription`. I checked out the implementation here: https://reasonml.github.io/en/try?rrjsx=true&reason=C4TwDgpgBAThCOBXCBnYAKA5AIwJRQF4AoKKAHygCEB7ADyzwG4ijRIoALAQwDsATADYQYWLgBooOCZgDG+YqQoAJXoIjp01MMACW1HqNzS8hAHySuuAFydVQkVt37Dx6ZZLkoAOWor+Qm25-YXQAfVdJJhY2aEQUCABlRGwUGRgdbT0DTHFI6TlCD3QAPyC1GEC7EJyI2SMoYrgkVGAbJuQ0Blx5c1lmIiFgKAAzampC0nQYqFzp7AlpmQlSqorbYJFc+ag5ZfaWtoQOjDxrHbMoAG8PUkGqOnQYeVgjluZSUhQAdx1gGQ4oOgyvZ8NcPp4-GogatQsMeD11uVYQYfDwIBInjdPD5IUILjAsQBfd5QYkeIijaglYHCAg4qp7V5oAg0egAVm6zEp1NWdN8DIa+2ZrPQACI2aLORSxjyNgRcep0FwZEsoHwuMBLBcwVBvr9-oDlQUdRRUdACOYANrqzUAXQ8FAS1AAtuoBAiBFAAAJQa0arj20iEoiE+qNJnAFkPDm4RhAA and I'll confess I don't 100% understand it, but it seems like an interesting idea. I guess the major discussions to have are:

1) Do we want to consolidate the API surface into 1 hook / 1 component or do 2 hooks / 2 components to handle `handler`?
2) If we go for 1 hook / 1 component can we simplify @gugahoa's implementation in any way or use @Schmavery's idea of a default `handler` that'd behave like `(~prevSubscriptions as _, ~subscription) => subscription`?
3) If we go for 2 hooks / 2 components, can we find a nice way to consolidate the shared pieces?

I'd say let's discuss things here and folks can feel free to submit PRs w/ suggested implementations. Thanks to you both so much for all these contributions too, I can really feel the lib getting better and better.

gugahoa:
[...]

PR #73, opened 2019-06-21
Use GADT handler type to enable polymorphic useSubscription

gugahoa:
Refer to comment https://github.com/FormidableLabs/reason-urql/issues/72#issuecomment-504251354 for more context

This PR encompass just enough so that a 1 hook / 1 component approach can be better discussed, the API can be a bit more polished so that `~handler` become optional


src/hooks/UrqlUseSubscriptionWithHandler.re
```diff
@@ -1,38 +1,42 @@
...
-external useSubscriptionWithHandlerJs:
+external useSubscriptionJs:
   (
-    useSubscriptionWithHandlerArgs,
-    UrqlTypes.parsedHandler('acc, 'response)
+    useSubscriptionArgs,
+    option((option('a), 'b) => 'a)
   ) =>
-  array(useSubscriptionWithHandlerResponseJs('acc)) =
+  array(useSubscriptionResponseJs('c)) =
```
gugahoa: This increases the type unsafety surface area, but it's safely wrapped in `useSubscription`.
In my opinion, the rationale going here can be better explained in comments if we choose this approach.

src/hooks/UrqlUseSubscriptionWithHandler.re
```diff
@@ -1,38 +1,42 @@
+type handler('a, 'b, 'c) =
```
parkerziegler: Reading through your comment in #72 (thanks so much for that), could we make these type arguments match the following naming convention?

```suggestion
type handler('acc, 'response, 'ret) =
```

That way it's super clear which type arguments apply to which cases.

src/hooks/UrqlUseSubscriptionWithHandler.re
```diff
@@ -1,38 +1,42 @@
...
-type useSubscriptionWithHandlerArgs = {
+type useSubscriptionArgs = {
   query: string,
   [@bs.optional]
   variables: Js.Json.t,
 };
 
 [@bs.deriving abstract]
-type useSubscriptionWithHandlerResponseJs('acc) = {
+type useSubscriptionResponseJs('a) = {
```
parkerziegler: I'd prefer to stick with `'acc` if you're ok with that. I'm more and more convinced that it's nice to have our type parameter names be meaningful for future readers of the code.

src/hooks/UrqlUseSubscriptionWithHandler.re
```diff
@@ -48,25 +52,28 @@ let useSubscriptionResponseToRecordWithHandler =
...
 
-let useSubscriptionWithHandler =
+let useSubscription =
     (
-      ~request: UrqlTypes.request('response),
-      ~handler: UrqlTypes.parsedHandler('acc, 'response),
-    ) => {
+      type a, type b, type c,
+      ~handler: handler(a, b, c),
+      request: UrqlTypes.request(b),
```
parkerziegler: Can you explain the logic around supplying `~handler` as a named function parameter and leaving `request` unnamed? Just not super familiar with this style 😅 and I'd love to understand the reasoning!

src/hooks/UrqlUseSubscriptionWithHandler.re
```diff
@@ -48,25 +52,28 @@ let useSubscriptionResponseToRecordWithHandler =
...
-    state |> useSubscriptionResponseToRecordWithHandler;
-  useSubscriptionResponse;
-};
\ No newline at end of file
+  let state: useSubscriptionResponse(c) = switch (handler) {
+    | NoHandler => 
+      useSubscriptionJs(args, None)[0] 
+      |> useSubscriptionResponseToRecord(parse)
+    | Handler(handler_fn) =>
+      useSubscriptionJs(args, Some(handler_fn))[0]
```
parkerziegler: So one thing I'm concerned about here – from what I can understand, the individual subscriptions coming in as the second argument to `handler` _won't_ get parsed properly in this implementation, which would break type inference. That was the idea behind `parsedHandler`. Basically, we would apply the user supplied `handler` to `prevSubscriptions` and `parse(subscription)`, such that each new `subscription` coming in would have its type inferred. Does this do the same? I.e. if I passed this request:

```
module SubscribeRandomInt = [%graphql
  {|
  subscription subscribeNumbers {
    newNumber @bsDecoder(fn: "string_of_int")
  }
|}
];
```

with this `handler`:

```
let handler = (~prevSubscriptions, ~subscription) => {
  switch (prevSubscriptions) {
  | Some(subs) => Array.append(subs, [|subscription|])
  | None => [|subscription|]
  };
};
```

I'd expect the type of `data` to be inferred properly as `array({. "newNumber": string })`. Does that happen in the example w/ this implementation?

src/hooks/UrqlUseSubscriptionWithHandler.re
gugahoa: Sure!

src/hooks/UrqlUseSubscriptionWithHandler.re
gugahoa: At first I went with handler as a named argument so that later we could make it optional, but after looking into it, it seems like we won't be able to make it optional so we can change it to a normal argument

src/hooks/UrqlUseSubscriptionWithHandler.re
gugahoa: Nice catch, it really isn't properly parsing the second argument before supplying it to handler. It wouldn't mess up type inference, but it does mess up type safety.

src/hooks/UrqlUseSubscriptionWithHandler.re
```diff
@@ -1,38 +1,42 @@
+type handler('acc, 'resp, 'ret) =
```
parkerziegler: Now that we have our GADT `handler` here, I think we can get rid of `handler` and `parsedHandler` in `UrqlTypes.re`, correct?

parkerziegler: Let's also make sure to rename this module to `UrqlUseSubscription.re` and get rid of the old `UrqlUseSubscription.re` before merging 😄

src/hooks/UrqlUseSubscriptionWithHandler.re
```diff
@@ -48,25 +52,28 @@ let useSubscriptionResponseToRecordWithHandler =
...
 
-let useSubscriptionWithHandler =
+let useSubscription =
     (
-      ~request: UrqlTypes.request('response),
-      ~handler: UrqlTypes.parsedHandler('acc, 'response),
-    ) => {
+      type acc, type resp, type ret,
+      ~handler: handler(acc, resp, ret),
+      request: UrqlTypes.request(resp),
```
parkerziegler: I think I'd still like to keep `request` as a labelled argument here if possible. It feels like a nicer API to have two labelled args rather than one labelled and one unlabelled. I'm forgetting, is there a specific reason we can't do this?

gugahoa: There's no issue in having only labeled arguments, only if all the labeled arguments are optional with `=?` it's required to have at least one non optional argument.

src/hooks/UrqlUseSubscriptionWithHandler.re
gugahoa: I was thinking of only doing this as the last step, so that the line diff doesn't become too big to review

For now I'm not reexporting `UrqlUseSubscription` instead of deleting it right away.

Do you think it's better to just delete it now?

parkerziegler: Good point, let's do it as the last commit after approval 👍

Schmavery:
Haven't been sure what thoughts to have on this. It's definitely some nice use of gadts @gugahoa, I just hope it won't throw people off (though the part that's exposed just looks like a regular variant to users so it's probably not too bad!).
It's too bad that it ends up making the simpler usecase require the NoHandler variant. The only thought I have right now is that if users of this api use the simple NoHandler version more often than not, maybe it's worth exposing that separately... I guess nothing is stopping the user or someone making a useSimpleSubscription that just calls useSubscription with NoHandler passed in. It's probably fine!

parkerziegler:
Good thoughts @Schmavery. I think I'm happy enough with this as an initial API. I am somewhat anxious to get v1 launched so people can start using it, and then we can iterate on this API once more feedback comes in.

src/hooks/UrqlUseSubscriptionWithHandler.re
gugahoa: When removing the `handler` and `parsedHandler`, I noticed that `UrqlSubscription` uses them. I think it's better to leave them here for now and tackle it in a new PR focused on migrating `UrqlSubscription` to use this approach

gugahoa:
@parkerziegler I believe all the concerns you raised have been properly addressed now, pinging of a re-review! :ok_hand:

PR #71, opened 2019-06-20
updated 1-execute-query-mutation

idkjs:
updated 1-execute-query-mutation to using useReducer hook

PR #74, opened 2019-06-21
updated 1-execute-query-mutation to using useReducer hook

idkjs:
Per https://github.com/FormidableLabs/reason-urql/pull/71#issuecomment-504480956

### DIFF
--- a/examples/1-execute-query-mutation/src/Previewer.re
+++ b/examples/1-execute-query-mutation/src/Previewer.re
@@ -93,7 +93,7 @@ module Styles = {
   let colors = {query: "48a9dc", mutation: "db4d3f"};
 };
 
-let str = ReasonReact.string;
+let str = React.string;
 
 type state = {
   query: string,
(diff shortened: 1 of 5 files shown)


examples/1-execute-query-mutation/src/Previewer.re
```diff
@@ -101,117 +101,108 @@ type state = {
...
+          ->str
+        </div>
+        <button
+          className={Styles.colors.query->Styles.button}
+          onClick={_event => executeQuery()}>
+          "Execute Query"->str
+        </button>
+      </section>
+      {switch (String.length(state.query)) {
+       | 0 => ReasonReact.null
```
parkerziegler: While we're here would you mind switching this to `React.null`? Just want to get rid of all `ReasonReact` references while this is fresh ❤️

idkjs: Also changed a few `ReasonReact.string` to `Reason.string`. Unless you didn't want that changed?

idkjs:
Done.

parkerziegler:
@all-contributors add @idkjs for code, example.

Issue #70, opened 2019-06-20
Implement better type inference in components.


parkerziegler:
Based off on work in #68, #69, and #60, we'll want to ensure our components have the same level of type safety as our hooks now do. This gets trickiest with `Subscription`, whose return data type will differ based on whether a `handler` is supplied or not. We may consider doing what we did w/ the hooks implementation where we have separate `Subscription` and `SubscriptionWithHandler` components to handle this polymorphism.

parkerziegler:
## Roadmap

- [ ] Query
- [ ] Query Examples
- [ ] Mutation
- [ ] Mutation Examples
- [ ] Subscription
- [ ] Subscription Examples

PR #77, opened 2019-06-23
Improve type inference on components.

parkerziegler:
This PR begins to address #70. Taking cue from the recent hooks PRs, the major change here is to introduce a type argument, `'response`, that matches the return value of calling `parse` on the data returned by `executeQuery` and `executeMutation`. `Query` and `Mutation` now also take a `request` prop in lieu of `query` and `variables`. `Subscription` will be done in a follow up since the hooks API w/ GADTs in still in flight in #73. 

Some noise in this diff is caused by `refmt` stuff, so I'll get a pre-commit hook setup w/ latest `refmt` to ensure we keep this consistent for everyone 👍

cc/ @gugahoa @Schmavery including you here as well for review if you have the time (no worries if not).

### DIFF
--- a/src/components/UrqlMutation.re
+++ b/src/components/UrqlMutation.re
@@ -47,7 +47,7 @@ let urqlDataToRecord = (parse, variables, result) => {
 let make =
     (
       ~request: UrqlTypes.request('response),
-      ~children: mutationRenderProps('a) => React.element,
+      ~children: mutationRenderProps('response) => React.element,
     ) => {
   let query = request##query;
   let variables = request##variables;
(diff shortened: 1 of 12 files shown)


src/components/UrqlMutation.re
```diff
@@ -39,18 +40,19 @@ let urqlDataToRecord = (result: mutationRenderPropsJs('a)) => {
...
-  };
+  {fetching, data, error, executeMutation, response};
 };
 
 [@react.component]
 let make =
-    (~query: string, ~children: mutationRenderProps('a) => React.element) =>
+    (
+      ~request: UrqlTypes.request('response),
+      ~children: mutationRenderProps('a) => React.element,
```
gugahoa: The inferred type here will be `'response`, but I believe it's still good to change `'a` to `'response` to be explicit
```suggestion
      ~children: mutationRenderProps('response) => React.element,
```

parkerziegler: Ah good catch, thought I got them all!

Issue #78, opened 2019-06-23
Add bsrefmt pre-commit hook with husky and lint-staged.


parkerziegler:
In order to ensure everyone is using the same version of `refmt` when committing their code, we'll want to add `husky` and `lint-staged` to run on pre-commit.

parkerziegler:
Closed by #79.

PR #79, opened 2019-06-24
Add pre-commit hook for refmt. Run refmt over codebase.

parkerziegler:
Addresses #78. This PR adds a pre-commit hook to `reason-urql` to run `refmt` over the codebase. We've had enough drift in PRs created by different editor setups that I figured it was time to move this into a pre-commit step. cc/ @gugahoa @Schmavery @kiraarghy let me know if you have any feelings here. I tend to like pre-commit hooks to keep diff noise down to meaningful changes, while still allowing folks to have their editors set up they way they like.


__tests__/UrqlClient_test.re
```diff
@@ -116,4 +116,4 @@ describe("UrqlClient", () => {
       )
     );
   });
-});
\ No newline at end of file
+});
```
parkerziegler: A _lot_ of these files just involved adding a carriage return to the file. I'm going to make sure that my local editor setup matches this project's `refmt` (I'm using `reason-language-server`, not sure if any of you happen to know if there's an easy way to check this).

PR #81, opened 2019-06-25
Add common hook type (Fix #80)

### LINKED ISSUE #80: Fix improper type inference in hooks.
cc/ @gugahoa @Schmavery I'm not totally sure how to fix this one, so reaching out for help.

In `examples/2-query` in `Monster.re`, change L37 from this:

```reason
let (result, _) = Hooks.useQuery(~request, ());
```

to this:

```reason
useQuery(~request, ());
```

and add this at the top:

```
open Hooks;
```

When doing this, I get an interesting type error:

```
We've found a bug for you!
  /Users/parkerziegler/Documents/repos/OSS/reason-urql/examples/2-query/src/Monster.re 38:25-46
  
  36 ┆    value passed in from GetAll */
  37 ┆ let request = GetPokemon.make(~name=pokemon, ());
  38 ┆ let ({response}, _) = useQuery(~request, ());
  39 ┆ 
  40 ┆ switch (result.response) {
  
  This has type:
    ReasonUrql.Hooks.useQueryResponse({. "pokemon": option({. "classification": 
                                                             option(string),
                                                             "height": 
                                                             option({. "maximum": 
                                                                    option(
                                                                    string)}),
                                                             "image": 
                                                             option(string),
                                                             "name": 
                                                             option(string),
                                                             "weight": 
                                                             option({. "maximum": 
                                                                    option(
                                                                    string)})})})
      (defined as
      (ReasonUrql.Hooks.useQueryState({. "pokemon": option({. "classification": 
                                                             option(string),
                                                             "height": 
                                                             option({. "maximum": 
                                                                    option(
                                                                    string)}),
                                                             "image": 
                                                             option(string),
                                                             "name": 
                                                             option(string),
                                                             "weight": 
                                                             option({. "maximum": 
                                                                    option(
                                                                    string)})})}),
      ReasonUrql.Hooks.partialOperationContextFn))
  But somewhere wanted:
    (ReasonUrql.Hooks.useSubscriptionResponse('a), 'b)
  
  The incompatible parts:
    ReasonUrql.Hooks.useQueryState({. "pokemon": option({. "classification": 
                                                          option(string),
                                                          "height": option(
                                                                    {. "maximum": 
                                                                    option(
                                                                    string)}),
                                                          "image": option(
                                                                   string),
                                                          "name": option(
                                                                  string),
                                                          "weight": option(
                                                                    {. "maximum": 
                                                                    option(
                                                                    string)})})})
      (defined as
      UrqlUseQuery.useQueryState({. "pokemon": option({. "classification": 
                                                        option(string),
                                                        "height": option(
                                                                  {. "maximum": 
                                                                    option(
                                                                    string)}),
                                                        "image": option(
                                                                 string),
                                                        "name": option(
                                                                string),
                                                        "weight": option(
                                                                  {. "maximum": 
                                                                    option(
                                                                    string)})})}))
    vs
    ReasonUrql.Hooks.useSubscriptionResponse('a) (defined as
      UrqlUseSubscription.useSubscriptionResponse('a))
```

I'm not totally sure why this is happening. I'm wondering if Reason is searching for the record type associated with the destrucutred `{ response }` and it's saying, "Hey, that record is `type useSubscriptionResponse`" That's my best guess. If so, I'm thinking we may need to do one-off `module UseQuery = UrqlUseQuery;` such that only the definitions from `UrqlUseQuery` are brought into scope and the proper record type `useQueryResponse` is inferred. Any other thoughts? I also tried adding a top-level `type hookReponse('ret)` since our hooks have consistent return structures, but I got an error w/ the `useSubscription` GADT b/c `type ret` there would escape its scope.

gugahoa:


Schmavery:
This is a super slick solution afaict, nice. Any thoughts about reexposing the result type (and any other relevant types) from the ReasonUrql.Hooks module so that users don't have to open another module?

src/UrqlTypes.re
```diff
@@ -100,3 +100,10 @@ type handler('acc) =
 
 type parsedHandler('acc, 'response) =
   (~prevSubscriptions: option('acc), ~subscription: 'response) => 'acc;
+
+type hookResponse('a) = {
```
parkerziegler: Nit – let's stick with `'ret` as the type parameter here to keep it consistent.

src/hooks/UrqlUseSubscription.re
```diff
@@ -18,20 +20,13 @@ type useSubscriptionResponseJs('ret) = {
...
-};
-
 [@bs.module "urql"]
 external useSubscriptionJs:
   (useSubscriptionArgs, option((option('acc), Js.Json.t) => 'acc)) =>
   array(useSubscriptionResponseJs('ret)) =
   "useSubscription";
 
-let useSubscriptionResponseToRecord = (parse, result) => {
+let useSubscriptionResponseToRecord = (parse, result): hookResponse('a) => {
```
parkerziegler: +1 to using `'ret` in lieu of `'a` here to make it easier to track type parameters.

src/hooks/UrqlUseSubscription.rei
```diff
@@ -2,16 +2,9 @@ type handler('acc, 'resp, 'ret) =
...
-  response: UrqlTypes.response('ret),
-};
-
 let useSubscription:
   (
     ~handler: handler('acc, 'resp, 'ret),
     ~request: UrqlTypes.request('resp)
   ) =>
-  useSubscriptionResponse('ret);
+  UrqlTypes.hookResponse('ret);
```
parkerziegler: Is there a particular reason we don't `open UrqlTypes` at the top in this file as well? I'm guessing it's because we still have those defs for `handler` hanging around, yeah? I'll remove those once we convert `Subscription` over 👍

gugahoa: There was no problem in opening `UrqlTypes` on that file, just changed it! Thanks

gugahoa:
I tried to make it not necessary to `open Types` when opening Hooks, but couldn't accomplish it, any ideas on how to do it?

I tried making a `type hookResponse('ret) = UrqlTypes.hookResponse('ret)` in each file, or putting it inside `module Hooks`

Pinging @parkerziegler for re-review

Schmavery:
I'm surprised that adding `type hookResponse('ret) = UrqlTypes.hookResponse('ret)` to the Hooks module didn't work. I was playing with it a little earlier, does this help?
https://reasonml.github.io/try?rrjsx=true&reason=LYewJgrgNgpgBAFQJ4AcYGc4F44G8BQccALqvAIbZ4CWAXHNQHbEC+A3ISWXAEbadEAPnABCIAB4AKJsQCUHdvnyhIsOAAkQIANaYcBIqTRxKOZGnQA6UzXozFh7nzNkrzuMLFSZ8pURXQ8ACCVAZEcLDEcABmVJK45hiW1Cyy2AB8DBxEipwBaiKhAhEwUbE48YlWKWlYmdTZcIq5IGiMGlq6HJEkoXRwAIyKPcQATFRekgO+QZbRknIcQA

Basically I reexposed the type using something like this in the Hooks module:
```reason
type a = Types.a = {i: int};
```

parkerziegler:
@gugahoa can you try the above suggested by @Schmavery? If we get that to work then I think this is set to 🚢

parkerziegler:
Another thing to bring up while I'm here – do you happen to know how to handle GADTs in the component case? I got it working-ish w/ `Subscription`, but this syntax will fail the build:

```
let make =
    (
      type acc,
      type response,
      type ret,
      ~request: request(ret),
      ~handler: handler(acc, response, ret),
      ~children: subscriptionRenderProps(ret) => React.element,
    ) => {
```

with the following error:

```
[1/2] Building src/components/UrqlSubscription.mlast
Fatal error: exception Invalid_argument("react.component calls can only be on function definitions or component wrappers (forwardRef, memo).")
ninja: error: rebuilding 'build.ninja': subcommand failed
error Command failed with exit code 2.
```

I wonder if there's any easy way around this 🤔

gugahoa:
@Schmavery thanks for the suggestion, it worked perfectly!
@parkerziegler I remember getting stuck on that error previously, but I don't remember what caused it. Is your work in any branch where we can take a look?

parkerziegler:
Yep, just pushed it up: https://github.com/FormidableLabs/reason-urql/tree/task/type-inference-gadt-for-subscription Here are the relevant lines: https://github.com/FormidableLabs/reason-urql/blob/task/type-inference-gadt-for-subscription/src/components/UrqlSubscription.re#L53-L80

gugahoa:
@parkerziegler seems like you found a bug, I tried to look for any issues that mentioned that, but couldn't find, so I created a new one on the reason-react repo: https://github.com/reasonml/reason-react/issues/452

parkerziegler:
Thanks @gugahoa. I'm skeptical that we'll get support for it in the near future, so I'll push up a possible solution for us so that we can feature cap `v1`.

PR #83, opened 2019-07-02
Fix ReasonUrql hooks and components interop safety

gugahoa:
Since we got closer to 1.0, I have been using reason-urql at my startup, and noticed some unexpected behaviors from the hooks.

## Problem 1
The response from JS urql can contain data that is either `null` or `undefined`, this is not properly covered by `[@bs.optional]`, the fix is to use `Js.Nullable.t('a)`, which can be safely converted to an `option('a)`

## Problem 2
The return value from reason-urql hooks were not safe to use in the dependency list of `useEffect`, `useMemo`, `useCallback` and similar, even though the response JS-land seems safe to use. That's caused by us converting the Js object to an Reason type, and at runtime that always creates a new object

### DIFF
--- a/src/components/UrqlMutation.re
+++ b/src/components/UrqlMutation.re
@@ -1,8 +1,7 @@
 [@bs.deriving abstract]
 type mutationRenderPropsJs = {
   fetching: bool,
-  [@bs.optional]
-  data: Js.Json.t,
+  data: Js.Nullable.t(Js.Json.t),
   [@bs.optional]
   error: UrqlCombinedError.t,
   executeMutation:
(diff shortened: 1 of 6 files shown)


gugahoa:
Found even more bugs, I'll update shortly with a description and fix in this same PR

parkerziegler:
Interesting, I think we had these types originally as `Js.Nullable.t`'s, but we converted over to `bs.optional` in the course of this PR: https://github.com/FormidableLabs/reason-urql/pull/59/files. If I'm understanding correctly, is the idea that, b/c `data` can potentially be _both_ `null` and `undefined`, `Js.Nullable.t` is the only safe option (whereas `@bs.optional` would only cover the `undefined` case).

In the second case, I'm also for using hooks and I like the approach a lot. Memoize the conversion from JS object structure to Reason record and only re-run on change of the JS object 🚀. I think that makes sense in that, as you described earlier, it prevents other hooks that rely on the response for shallow equality comparisons from re-computing memoized values (or rerunning in the case of `useEffect`). Will do a more thorough review in the morning. cc/ @Schmavery any thoughts?

Schmavery:
look pretty good to me, nice clean fixes :)

parkerziegler:
🚀 Thanks again @gugahoa! I can pick up `Query` today. If anyone else wants to tackle `Subscription` before I can get to it, feel free.

PR #84, opened 2019-07-03
Fix executeQuery signature

gugahoa:
A signature with `~opts: option(UrqlTypes.partialOperationContext)=?` will expect `opts` of type `option(option(UrqlTypes.partialOperationContext))`, instead of the intended (I believe) `option(UrqlTypes.partialOperationContext)`

### DIFF
--- a/src/UrqlClient.re
+++ b/src/UrqlClient.re
@@ -87,7 +87,7 @@ external executeQuery:
   (
     ~client: t,
     ~query: UrqlTypes.graphqlRequest,
-    ~opts: option(UrqlTypes.partialOperationContext)=?,
+    ~opts: UrqlTypes.partialOperationContext=?,
     unit
   ) =>
   Wonka.Types.sourceT('a) =


parkerziegler:
@gugahoa I wouldn't be surprised if this was true, but it does conflict with this (admittedly old) resource: https://2ality.com/2017/12/functions-reasonml.html#type-annotations-for-optional-parameters, which states that optional labelled args have to be wrapped with an `option` type when the type of the arg is explicitly written. I suppose this is something I can verify myself by playing with the examples, but just wanted to see if you know of any more recent resources that would support these changes? Wouldn't be surprised if Reason v3 syntax fixed this.

gugahoa:
Oh, it doesn't conflict:

> When you annotate optional parameters, they must all have option(···) types:
> ```reason
> let add = (~x: option(int)=?, ~y: option(int)=?, ()) =>
>   ···
> ```
> The type signature of add is:
> 
> ```reason
> (~x: int=?, ~y: int=?, unit) => int
> ```

In this case, when you're writing the actual function, you have to wrap with the `~x: option('a)=?`, but the signature of said function will be `~x: 'a=?`  
It's a little bit confusing

In this case, since the external directive is just a function signature for a FFI, it should use the `~x: 'a=?` signature for optional argument

parkerziegler:
Gotcha, thanks for helping to clarify that! There may be other places in the codebase with that sort of improper typing (additional `option`), so I'll be on the lookout as I continue to refactor things.

PR #85, opened 2019-07-05
Type cleanup in UrqlTypes and UrqlClient.

parkerziegler:
This PR attempts to do a final(ish) type pass on `UrqlTypes` and `UrqlClient`. I think we may still need to alter the APIs for `.executeQuery`, `.executeMutation`, and `.executeSubscription` on the `client` to properly infer the type of the response. We'll probably want to do something similar to what we did for the components and hooks by passing the whole request built by `graphql_ppx`, but that's a separate follow up PR. I'm considering releasing a `v1.0.0-beta.1` after these changes get in so we can give folks the hooks publicly while I get the type inference going on these methods 😄

cc/ @gugahoa @Schmavery

### DIFF
--- a/examples/1-execute-query-mutation/package.json
+++ b/examples/1-execute-query-mutation/package.json
@@ -6,6 +6,7 @@
   "author": "Parker Ziegler <parker.ziegler@formidable.com>",
   "license": "MIT",
   "scripts": {
+    "build": "bsb -make-world",
     "start": "bsb -make-world -w",
     "clean": "bsb -clean-world",
     "start:demo": "webpack-dev-server --hot"
@@ -16,7 +17,7 @@
     "reason-urql": "link:../../"
   },
   "devDependencies": {
-    "bs-platform": "^5.0.3",
+    "bs-platform": "5.0.4",
     "graphql_ppx": "^0.2.8",
     "webpack": "^4.29.6",
     "webpack-cli": "^3.2.3",
--- a/examples/1-execute-query-mutation/src/App.re
+++ b/examples/1-execute-query-mutation/src/App.re
@@ -1,18 +1,16 @@
-module Styles = {
-  open Css;
-
-  let page =
-    style([
-      position(absolute),
-      top(px(0)),
-      bottom(px(0)),
-      left(px(0)),
-      right(px(0)),
-      padding(px(20)),
-    ]);
-};
-
 [@react.component]
 let make = () => {
-  <div className=Styles.page> <ColdStart /> <Previewer /> </div>;
+  <div
+    style={ReactDOMRe.Style.make(
+      ~position="absolute",
+      ~top="0px",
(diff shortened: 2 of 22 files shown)


examples/1-execute-query-mutation/package.json
```diff
@@ -16,7 +17,7 @@
     "reason-urql": "link:../../"
   },
   "devDependencies": {
-    "bs-platform": "^5.0.3",
+    "bs-platform": "5.0.4",
```
parkerziegler: I decided to pin everything to `bs-platform 5.0.4` b/c I was getting errors about the root project using a different version of `bs-platform` than the example project. So keeping everything pinned to 5.0.4 now, but will keep upgrading. It's just a devDep, so I don't think this will affect users at all.

examples/1-execute-query-mutation/src/App.re
```diff
@@ -1,18 +1,16 @@
...
-      left(px(0)),
-      right(px(0)),
-      padding(px(20)),
-    ]);
-};
-
 [@react.component]
 let make = () => {
-  <div className=Styles.page> <ColdStart /> <Previewer /> </div>;
+  <div
```
parkerziegler: All the style changes basically boil down to 1) If it's a one-off style, just use `ReactDOMRe.Style`, and 2) If the styles are more extensive, use `bs-css` with a separate `module` for styles.

src/UrqlClient.re
```diff
@@ -108,58 +108,40 @@ external executeMutation:
...
-    ) => {
-  let type_ = UrqlTypes.operationTypeToJs(operationType);
-  createRequestOperationJs(
-    ~client,
-    ~operationType=type_,
-    ~request,
-    ~opts,
-    (),
-  );
+let createRequestOperation = (~client, ~operationType, ~request, ~opts, ()) => {
```
gugahoa: In the previous version `~opts` is an optional argument, making it non-optional is intended or a typo? If intended, you can also delete the last `()`

src/UrqlTypes.re
```diff
@@ -72,19 +78,16 @@ type operation = {
...
 
-/* The result of a GraphQL operation. */
-type operationData;
-
 [@bs.deriving abstract]
 type operationResult = {
   operation,
   [@bs.optional]
-  data: operationData,
+  data: Js.Json.t,
```
gugahoa: Should this be an `Js.Nullable.t(Js.Json.t)`?

src/UrqlClient.re
parkerziegler: Not intentional, good catch!

parkerziegler:
@gugahoa feedback addressed, let me know if there are any other changes you'd like to see. Planning to merge this and cut a `v1.0.0-beta.1` release today unless there are any other issues you see!

PR #88, opened 2019-07-14
Properly send fetchOption fn to urql Client

gugahoa:
When sending a `FetchFn` as `fetchOptions`, it was being called during the `Client` instantiation, but the expected behavior would be for it to be called on demand, as it is on upstream `urql` lib

This fix makes it possible to properly send a function as a `fetchOptions`, but introduce type unsafety with `Obj.magic` being called at `unwrapFetchOptions` where the type checker will just trust the user type annotation when calling it with a `None` value

To properly address that issue I believe the most sensible would be to not expose `unwrapFetchOptions` to any user. Any thoughts?

I'm marking this as WIP until that issue above is properly discussed

### DIFF
--- a/src/UrqlClient.re
+++ b/src/UrqlClient.re
@@ -2,14 +2,16 @@ open UrqlTypes;
 type t;
 
 /* Helpers for supporting polymorphic fetchOptions. */
-type fetchOptions =
-  | FetchOpts(Fetch.requestInit)
-  | FetchFn(unit => Fetch.requestInit);
+type fetchOptions('a) =
+  | FetchOpts(Fetch.requestInit): fetchOptions(Fetch.requestInit)
+  | FetchFn(unit => Fetch.requestInit)
+    : fetchOptions(unit => Fetch.requestInit);
 
-let unwrapFetchOptions = fetchOptions =>
+let unwrapFetchOptions = (type a, fetchOptions: option(fetchOptions(a))): a =>
   switch (fetchOptions) {
-  | FetchOpts(opts) => opts
-  | FetchFn(fn) => fn()
+  | Some(FetchOpts(opts)) => opts
+  | Some(FetchFn(fn)) => fn
+  | None => Obj.magic(Fetch.RequestInit.make())
   };
 
 /* A module for binding exchange types and urql's exposed exchanges. Since this module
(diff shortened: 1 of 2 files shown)


src/UrqlClient.re
```diff
@@ -2,14 +2,16 @@ open UrqlTypes;
 type t;
 
 /* Helpers for supporting polymorphic fetchOptions. */
-type fetchOptions =
-  | FetchOpts(Fetch.requestInit)
-  | FetchFn(unit => Fetch.requestInit);
+type fetchOptions('a) =
```
parkerziegler: 🆒this is another instance of the GADT approach, correct? I definitely liked that for `useSubscription`

src/UrqlClient.re
```diff
@@ -2,14 +2,16 @@ open UrqlTypes;
...
+    : fetchOptions(unit => Fetch.requestInit);
 
-let unwrapFetchOptions = fetchOptions =>
+let unwrapFetchOptions = (type a, fetchOptions: option(fetchOptions(a))): a =>
   switch (fetchOptions) {
-  | FetchOpts(opts) => opts
-  | FetchFn(fn) => fn()
+  | Some(FetchOpts(opts)) => opts
+  | Some(FetchFn(fn)) => fn
+  | None => Obj.magic(Fetch.RequestInit.make())
```
parkerziegler: Is there potential we could return an `option` type here to prevent the need for `Obj.magic`, i.e.:

```reason
let unwrapFetchOptions = (type a, fetchOptions: option(fetchOptions(a))): option(a) =>
   switch (fetchOptions) {
   | Some(FetchOpts(opts)) => Some(opts)
   | Some(FetchFn(fn)) => Some(fn)
   | None => None
   }
```

IIRC, BuckleScript will pass along the value wrapped by `Some` to the JS side, and in the `None` case will pass `undefined`, which should be totally fine for the `urql` side. I don't mind `Obj.magic` too much, so also ok with this implementation if you think it's better 👍

src/UrqlClient.re
gugahoa: It is! :blush:

src/UrqlClient.re
gugahoa: Great suggestion, I made this change, now we're 100% safe :^)

gugahoa:
@parkerziegler thanks for the review! I made the suggested change. Didn't know we could pass undefined to `urql` and it would be properly handled :ok_hand:

parkerziegler:
Released in v1.0.0-beta.2.

PR #89, opened 2019-07-18
Improve combinedError API. Bind using OCaml classes.

parkerziegler:
This PR adds nicer bindings for `urql`'s `CombinedError` API. The previous binding were just wrong, so you couldn't actually access anything on the `Error(e)` variant. Now, `e` will come back with fields for `networkError`, `graphqlErrors`, and `response`. These fields are accessible using `##`, i.e.:

```reason
switch (e##networkError) {
| Some(exn) => /* do something with exn of type Js.Exn.t */
| None => /* no network error occurred */
}
```

I added an example of the above to `Monster.re` in `examples/2-query`.

190718_combined_error_demo

Since `CombinedError` is a JS class instance on the `urql` side, I opted to use OCaml classes to bind it as described here: https://bucklescript.github.io/docs/en/class#bind-to-js-classes

@gugahoa let me know if this sounds like a decent enough API to you! I also thought about trying to do some error conversion to turn this into a record type, but wanted to start small and get thoughts first.

### DIFF
--- a/src/utils/UrqlCombinedError.re
+++ b/src/utils/UrqlCombinedError.re
@@ -54,9 +54,23 @@ class type _combinedError =
   {
     pub networkError: Js.Nullable.t(Js.Exn.t);
     pub graphqlErrors: Js.Nullable.t(array(graphqlError));
-    pub response: 'response;
+    pub response: Js.Nullable.t(Fetch.response);
   };
 
 type t = Js.t(_combinedError);
+
+type combinedError = {
+  networkError: option(Js.Exn.t),
+  graphqlErrors: option(array(graphqlError)),
+  response: option(Fetch.response),
+};
+
+let combinedErrorToRecord = (err: t): combinedError => {
+  {
+    networkError: err##networkError->Js.Nullable.toOption,
+    graphqlErrors: err##graphqlErrors->Js.Nullable.toOption,
+    response: err##response->Js.Nullable.toOption,
+  };
+};
 
 [@bs.module "urql"] external combinedError: t = "CombinedError";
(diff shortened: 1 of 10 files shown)


src/utils/UrqlCombinedError.re
```diff
@@ -39,24 +37,26 @@ type extension;
...
-  "networkError": Js.Nullable.t(Js.Exn.t),
-  "graphqlErrors": Js.Nullable.t(array(graphqlError)),
-  "response": Js.Nullable.t('a),
-};
+class type _combinedError =
+  [@bs]
+  {
+    pub networkError: Js.Nullable.t(Js.Exn.t);
+    pub graphqlErrors: Js.Nullable.t(array(graphqlError));
+    pub response: 'response;
```
gugahoa: What type is response? Also, should it be `Js.Nullable.t('response)`?

Is this type the same as the `'response` in the graphql request? As it is right now it could lead to type safety issues as we can't constrain it to any type

parkerziegler: Yep, good call out. I'll investigate what that actually is on the `urql` side and see if we can type it properly.

parkerziegler: Ok, so the proper type for it ends up being `Js.Nullable.t(Fetch.response)` b/c what you get back from `urql` is the literal `fetch` response object or `undefined`. Happy we get to use `bs-fetch` for the API 🎉 Another thing I did was add `combinedErrorToRecord`. This basically converts `urql`'s `combinedError` to a record type for us, which gives us a slightly nicer API in my opinion (`e.networkError`, `e.graphqlErrors`, `e.response`). Let me your thoughts @gugahoa and @Schmavery!

parkerziegler:
Planning to release `v1.0.0-beta.2` with this and the past few PRs either tonight or tomorrow morning so we can keep working out any kinks! Thanks again everyone 😄

PR #90, opened 2019-07-31
Add docs for v1.

parkerziegler:
This PR is our staging ground for adding all docs for `v1`.

PR #92, opened 2019-08-03
Add proper type inference to execute* methods on client.

parkerziegler:
Fix #90. 

As @Schmavery pointed out in that issue, the `client`'s `execute*` methods weren't correctly inferring the types of the response data. Part of the reason was that we were just directly using `Request.createRequest`, which didn't forward along the `parse` function from the `graphql_ppx` query module (the secret sauce that allows us to infer types properly).

This PR fixes this by changing up the API a bit. `Client.execute*` methods now accept the full `graphql_ppx` query module (or a `Js.t` that fulfills the same contract if not using `graphql_ppx`), which allows us to use that `parse` function to properly infer types. In a lot of ways this mimics what we do in our hooks and components 😄

cc/ @gugahoa including you as well – let me know what you think of this. @Schmavery if you feel good about this I'll make sure to add you as a co-author on the commit since your sample was the main inspiration for these changes.

I'll happily publish another beta release right after this gets merged so folks can pick it up.

### DIFF
--- a/examples/1-execute-query-mutation/src/Previewer.re
+++ b/examples/1-execute-query-mutation/src/Previewer.re
@@ -1,4 +1,5 @@
 open ReasonUrql;
+open Types;
 
 let client = Client.make(~url="https://formidadog-ql.now.sh", ());
 
@@ -60,18 +61,36 @@ let make = () => {
   let executeQuery = () =>
     Client.executeQuery(~client, ~request=queryRequest, ())
     |> Wonka.subscribe((. data) =>
-         switch (Js.Json.stringifyAny(data)) {
-         | Some(d) => dispatch(SetQuery(d))
-         | None => ()
+         switch (data.response) {
+         | Data(d) =>
+           switch (Js.Json.stringifyAny(d)) {
+           | Some(s) => dispatch(SetQuery(s))
+           | None => ()
+           }
+         | Error(e) =>
+           switch (Js.Json.stringifyAny(e)) {
+           | Some(s) => dispatch(SetQuery(s))
+           | None => ()
+           }
+         | _ => ()
          }
        );
 
   let executeMutation = () =>
     Client.executeMutation(~client, ~request=mutationRequest, ())
     |> Wonka.subscribe((. data) =>
-         switch (Js.Json.stringifyAny(data)) {
-         | Some(d) => dispatch(SetMutation(d))
-         | None => ()
+         switch (data.response) {
+         | Data(d) =>
+           switch (Js.Json.stringifyAny(d)) {
+           | Some(s) => dispatch(SetQuery(s))
+           | None => ()
+           }
+         | Error(e) =>
+           switch (Js.Json.stringifyAny(e)) {
+           | Some(s) => dispatch(SetQuery(s))
+           | None => ()
+           }
+         | _ => ()
          }
        );
 
--- a/examples/1-execute-query-mutation/src/Previewer.re
+++ b/examples/1-execute-query-mutation/src/Previewer.re
@@ -14,7 +14,7 @@ module GetAllDogs = [%graphql
 |}
 ];
 
-let queryRequest = Request.createRequest(~query=GetAllDogs.make()##query, ());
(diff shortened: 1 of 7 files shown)


examples/1-execute-query-mutation/src/Previewer.re
```diff
@@ -93,7 +91,7 @@ let make = () => {
         </div>
         <button
           className={PreviewerStyles.colors.query->PreviewerStyles.button}
-          onClick={_event => executeQuery()}>
+          onClick={_event => executeQuery() |> ignore}>
```
parkerziegler: These ignores had to get added b/c `.subscribe` returns one of `wonka`'s `subscriptionT` types (basically just allows you to unsubscribe from the query).

src/UrqlClient.re
```diff
@@ -87,37 +87,97 @@ type clientOptions('a) = {
...
+  let req =
+    UrqlRequest.createRequest(
+      ~query=request##query,
+      ~variables=request##variables,
+      (),
+    );
+  let parse = request##parse;
+
+  urqlExecuteMutation(~client, ~mutation=req, ~opts?, ())
+  |> Wonka.map((. res) => parse(res##data));
```
parkerziegler: `map` seemed like the most logical operator b/c we're applying `parse` as a transformation of `res##data`, such that what consumers actually get matches the type signature returned by `parse`.

Schmavery:
One thing I left out of my example accidentally was the little wrapper around the parse function that turns it into the same variant as is returned by the hooks.. I think this might fail sometimes if parse gets called with an empty `res##data`?

Something like 
```reason
type response('a) = | Fetching | NotFound | Data('a) | Error(combinedErrorThingy);

[@bs.deriving abstract]
type responseJs = {
  fetching: bool,
  data: Js.nullable.t(Js.Json.t),
  [@bs.optional]
  error: ???
};

let processResponse = (parseFn, response) => {
  let data = response->dataGet->Js.Nullable.toOption->Belt.Option.map(parse);
  let error = response->errorGet;
  let fetching = response-fetchingGet;

  switch (fetching, data, error) {
  | (true, _, _) => Fetching
  | (false, Some(data), _) => Data(data)
  | (false, _, Some(err)) => Error(err)
  | _ => NotFound
};
```
And then like here: https://github.com/FormidableLabs/reason-urql/pull/92/files/85350495a33cfe78f7a1a2ccb83ea7bf3ec67fce#diff-b5caf26092fde2e2f066a18fa8a6a7f5R145

Doing something like this:
```reason
let parse = processResponse(request##parse);
```

Obvs feel free to play with the types, some of it is just copy pasted from the hooks/components code. I was just worried that there may be more states than just having valid data returned, but what do I know, maybe those errors have already been filtered out by this point! haha thanks for taking such a quick look at this!

parkerziegler:
@Schmavery good point – that you're right that we'll need error handling support, and I think the signature of the response should be exactly the same as the one we're using for components and hooks. I'll try to get this sorted later tonight or tomorrow. I'd kind of like to reconcile that `urqlDataToRecord` function in another module so we don't keep re-defining it.

parkerziegler:
Alright, I think I got this working @Schmavery. The key issue this exposed for me is that we need more isolated types – it's very rare that we actually want something in `UrqlTypes` b/c we have a lot of data structures that are _slightly_ different from one another. In particular, we run into issues with `[@bs.deriving abstract]`s that have the same keys, meaning we get clashes in the generated accessors. I'm thinking a nicer architecture could be for each module to have a `module Types` within it, i.e. `UrqlClient.Types`, `UrqlQuery.Types`, `UrqlUseQuery.Types`. The `Client` is probably the most important one b/c it exposes so many types used by other `module`s. In any case, I think we can merge this for now and I'll make an issue containing the above details.

Schmavery:
Yeah good idea. I'd be interested to see how different those types are.

parkerziegler:
They're not terribly different, but because we're doing a lot of conversions to records I think we need to keep them separate. I'm going to merge this for now, and work on a follow up to better isolate the types.

PR #93, opened 2019-08-09
Pin urql version to 1.0.5

Schmavery:
Urql 1.3.0 has a breaking change in the useMutation hook.
(Will fill this in with more information later but basically right now the useMutation doesn't do anything, will probably repro on the example project).
Discovered during reason-dojo :)

Edit: I think that the issue miiight be that https://github.com/FormidableLabs/urql/pull/351 added a second arg to the executeMutation cb, and then bsc curries the application of the first arg (cause the bindings only think the cb takes one arg) and results in the executeMutation cb never actually getting called/executing.

Co-authored-by: @AryehAlexander
Co-authored-by: @figitaki

### DIFF
--- a/__tests__/__snapshots__/UrqlClient_test.bs.js.snap
+++ b/__tests__/__snapshots__/UrqlClient_test.bs.js.snap
@@ -14,7 +14,6 @@ Client {
   "operations$": [Function],
   "reexecuteOperation": [Function],
   "results$": [Function],
-  "suspense": false,
   "url": "https://localhost:3000",
 }
 `;
@@ -33,7 +32,6 @@ Client {
   "operations$": [Function],
   "reexecuteOperation": [Function],
   "results$": [Function],
-  "suspense": false,
   "url": "https://localhost:3000",
 }
 `;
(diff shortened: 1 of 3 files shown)


parkerziegler:
Thanks @Schmavery! Would you mind taking the following steps?

1. Pin to `urql@1.0.5`. I think this is the best version to match for where the bindings are at currently.
2. Run `yarn` to regenerate `yarn`'s lock file.

Sorry that 1.3.0 introduced the breaking change for the bindings, but good to know that moving down to 1.2.0 fixed it. I'm trying to be a bit more cautious with version pinning since `urql`'s internals are evolving fast, so we'll pin to 1.0.5 for now, then migrate to 1.1 when we add in SSR support.

Schmavery:
Done :)

parkerziegler:
Mind just updating the snapshots? They're failing simply because the `suspense` flag lives on the client, but didn't get added until `1.1.0`. A simple `yarn test -u` should fix it, and then just commit the updated snapshots.

Schmavery:
no prob, looks like they're passing now

PR #97, opened 2019-09-07
Move Error branch upward

gugahoa:
During the conversion from JS response to record, the `Data` branch is checked before the `Error` branch, but it's to my understanding that there can't be `Data` when the server returns a graphql error (Instead of an error in the schema)

### DIFF
--- a/src/components/UrqlMutation.re
+++ b/src/components/UrqlMutation.re
@@ -43,8 +43,8 @@ let urqlMutationResponseToReason =
   let response =
     switch (fetching, data, error) {
     | (true, _, _) => UrqlTypes.Fetching
-    | (false, Some(data), _) => Data(data)
     | (false, _, Some(error)) => Error(error)
+    | (false, Some(data), _) => Data(data)
     | (false, None, None) => NotFound
     };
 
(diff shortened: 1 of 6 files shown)


Schmavery:
Are you sometimes getting both error and data? Just curious

gugahoa:
Yes, there's the case of this response: 
```{
  "data": {
    "login": null
  },
  "errors": [
    {
      "locations": [
        {
          "column": 0,
          "line": 2
        }
      ],
      "message": "invalid user-identifier",
      "path": [
        "login"
      ]
    }
  ]
}```

the login query returns a nullable session

parkerziegler:
Thanks @gugahoa and sorry for the late reply – busy getting ready for a conf talk on Reason this week! This makes sense to me, the JS equivalent is to check `error` before `data`, so I think this switch in the pattern matching makes more sense ✅ Errors should be higher precedence than data. I'd be curious to hear your thoughts on an `ErrorAndData(e, d)` variant that'd capture the case where you have `(Some(e), Some(d))`. Merging this as is for now and will release later this week when I get more breathing room (likely Saturday / Sunday).

gugahoa:
Good luck on your talk! :)
After reading a bit, I believe there can't be an `ErrorAndData`, and the issue with the case I got is because of some assumption somewhere in the codebase that assumes that if data is not null or undefined, it has meaningful value, but when getting a graphql/network error the content of data will be `data = { query_name: null }` if the return value of `query_name` is nullable

Although what I said above can be my own assumption being wrong too, I'll investigate this and research a bit before making a new issue to address this case. What do you think?

PR #99, opened 2019-09-15
Rename Types module to UrqlTypes to prevent namespace collisions. Docs cleanup. Upgrade dependencies.

### LINKED ISSUE #98: Namespace collisions
You guys have a module called Types on `ReasonUrql` which will/does clash with anyone who defines a module called types in their project. Just an FYI. Please let me know if you need further explanation. 

For future reference for readers. You can get around this by add `namespace:true` to your `bsconfig.json` and referencing your own types module with `open MyProjectNamespace.Types;` or `MyProjectNamespace.Types.myfunction();`

parkerziegler:
This PR fixes #98 and is targeted to be the last merged code before releasing stable v1. The list of changes includes:

- Preserving `UrqlTypes` as the name for the `module` containing library level types as it's fairly easy for this to conflict with a user-defined `Types` `module`, which feels at least somewhat standard for larger Reason projects.
- Upgrade dependencies, specifically `wonka` to latest (^3.1.0). 3.1.0 fixed a subtle issue around namespacing that was released in 3.0.0, hence the caret pinning to that version. See more here: https://github.com/kitten/wonka/pull/32
- An expansion of the docs to include a section on exchanges and editing some of the example code to be more useful to library users.
- Fix `examples/3-mutation`, which I forgot to upgrade when introducing changes around client `execute*` methods and type inference.
- Upgrade the `graphql_schema.json` for examples that use the `formidadog-ql` GraphQL API.

cc/ @Schmavery @gugahoa @idkjs My plan is to merge this if no one has objections, then cut a stable v1 so we can begin work on bindings for SSR, the expanded `urql` API, and `urql-exchange-graphcache` and the `urql-devtools-exchange`. Beyond that point, all future releases will follow semantic versioning. My inclination is to ramp up the `urql` version we bind to gradually as API features get introduced, otherwise I think we risk having a little too much ground to cover – let me know any thoughts on this end ❤️


idkjs:
Just re-ran https://github.com/FormidableLabs/reason-urql/tree/master/examples/3-mutation off of branch `task/v1-rc.0`. Still picking up Types from [`ReasonUrql.UrqlClient.Types`](https://github.com/FormidableLabs/reason-urql/blob/3213402b6afea476b819116ca663de3b68067311/src/UrqlClient.re#L18) when opening local `Types`:

Screenshot 2019-09-16 at 9 00 20 AM

Its working here https://github.com/idkjs/reason-urql/tree/task/v1-rc.0

parkerziegler:
Ah yep, I just missed that one, will amend. I'm guessing the compiler will pick up your local `Types.re` if you don't `open Client`, is that correct? Regardless, definitely want to ensure no user is getting these sorts of conflicts from something we can take care of on our end. Thanks @idkjs!

idkjs:
I did not check for that but yeah, not opening `Client` wont bring that `UrqlTypes.Types` into scope, right? Or am i talking out the side of my neck?

PR #100, opened 2019-09-16
Fix UrqlClient.Types -> UrqlClient.ClientTypes

idkjs:
I randomly chose a name for `UrqlClient.ClientTypes`. If there is a better one that conforms with what you are doing....

Test are failing:

```sh
   Details:

    /Users/prisc_000/Downloads/idkjs-reason-urq/node_modules/bs-fetch/src/Fetch.js:3
    import * as Curry from "bs-platform/lib/es6/curry.js";
           ^

    SyntaxError: Unexpected token *

       6 | var Block = require("bs-platform/lib/js/block.js");
       7 | var Curry = require("bs-platform/lib/js/curry.js");
    >  8 | var Fetch = require("bs-fetch/src/Fetch.js");
         |             ^
       9 | var ReasonUrql = require("../src/ReasonUrql.bs.js");
      10 |
      11 | Jest.describe("UrqlClient", (function (param) {

      at ScriptTransformer._transformAndBuildScript (node_modules/@jest/transform/build/ScriptTransformer.js:471:17)
      at ScriptTransformer.transform (node_modules/@jest/transform/build/ScriptTransformer.js:513:25)
      at Object.require (__tests__/UrqlClient_test.bs.js:8:13)
```


examples/3-mutation/package.json
```diff
@@ -19,7 +19,7 @@
   "devDependencies": {
     "bs-platform": "5.0.4",
     "graphql_ppx": "^0.2.8",
-    "webpack": "^4.29.6",
+    "webpack": "4.40.2",
```
parkerziegler: Any reason for this particular change? I'd just as soon leave this out for now, unless it's absolutely critical to the example running.

examples/3-mutation/src/Grid.re
```diff
@@ -1,17 +1,17 @@
...
-  name: string,
-  breed: string,
-  likes: int,
-  pats: int,
-  treats: int,
-  bellyscratches: int,
-  description: string,
-  imageUrl: string,
-};
+open Types;
```
parkerziegler: I'd prefer to keep the `type dog` local here just since it's used for the `@bsDecoder` stuff from `graphql_ppx` and it keeps it all on one place.

idkjs: It was just to test it. Removed.

examples/3-mutation/package.json
idkjs: got an error when running so re-installed. Let me check it.

idkjs: back to `"webpack": "^4.29.6"`

idkjs:
Thank you, sir. Looking forward to the bindings for https://github.com/FormidableLabs/urql-exchange-graphcache. You all don't have something started that I can hack on for that, do you?

parkerziegler:
@idkjs We don't currently, but I'll get this setup soon. I'm headed to London tomorrow and @kitten and I should have some time to chat more about our Reason bindings ✅ This is definitely on the roadmap, but there are some other features we need to add support for first.

parkerziegler:
That being said, we'll definitely get a repo setup for hacking sometime this week!

PR #104, opened 2019-09-23
Make bs-fetch a peerDependency.

parkerziegler:
This PR attempts to make `bs-fetch` a `peerDependency` for `reason-urql`. `bs-fetch` is solely used to type a few fields on the `Client` and the `combinedError`, but otherwise isn't actually used. It needs to be there to support the compiler in these modules, but in order to prevent our dep from conflicting with a user's own version of `bs-fetch`, we don't want to actually ship our dependency on `bs-fetch`.

cc/ @gugahoa I haven't yet ported this over to the examples, but let me know if you're able to apply this patch to see if it can help fix your issues! I think it'd be a win for the lib overall if we could remove this dependency. Will follow up on the examples tomorrow, just a bit late here in London!

### DIFF
--- a/bsconfig.json
+++ b/bsconfig.json
@@ -23,7 +23,7 @@
   ],
   "suffix": ".bs.js",
   "namespace": false,
-  "bs-dependencies": ["reason-react", "bs-fetch", "wonka"],
+  "bs-dependencies": ["reason-react", "wonka", "bs-fetch"],
   "bs-dev-dependencies": ["@glennsl/bs-jest"],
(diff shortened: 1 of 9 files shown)


parkerziegler:
Sweet, glad it worked @gugahoa! Looks like the build just passed. I'll get this merged and publish v1.0.1 with this change. Thanks!

parkerziegler:
Released in [v1.0.1](https://github.com/FormidableLabs/reason-urql/releases/tag/v1.0.1).

PR #105, opened 2019-09-26
Expose message field on combinedError

Schmavery:
Hacked-together PR to expose the message field from the urql combinederror object https://github.com/FormidableLabs/urql/blob/6e0e6557b487bc2c416215a9c92cce809b8bf9df/src/utils/error.ts#L64

Might still need to test this out but I wanted to get someone's opinion on the idea.

A couple times I've wanted to just quickly log the error without needing to handle all the cases/formatting etc. In the past I've made the mistake of  just `Js.String.make`ing the error, but that just gives some `[object Object]`-style output because urql defines the toString on their error object but we convert it to a record, which doesn't have the custom toString unfortunately. I looked into exposing the toString for the record version, but then realized that all the toString does is return the `message` field anyway so we might as well expose that directly?

### DIFF
--- a/src/utils/UrqlCombinedError.re
+++ b/src/utils/UrqlCombinedError.re
@@ -55,6 +55,7 @@ class type _combinedError =
     pub networkError: Js.Nullable.t(Js.Exn.t);
     pub graphqlErrors: Js.Nullable.t(array(graphqlError));
     pub response: Js.Nullable.t(Fetch.response);
+    pub message: string;
   };
 
 type t = Js.t(_combinedError);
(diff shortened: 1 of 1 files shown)


parkerziegler:
This makes a lot of sense to me @Schmavery. Honestly, I think `combinedError` is the weakest part of the bindings right now and some more effort on my part to document it and explore it is definitely needed. Let me play with this patch a bit, but I think this should work just fine!

parkerziegler:
Thanks for this @Schmavery! It definitely makes sense to add this in my opinion – it'll be a nicer mechanism for a lot of folks to deal with when they're just hacking around and need to get some error messaging. Applied the patch locally and tested against a few examples – everything works great! I'll add it in v1.0.1 (going out in the next few mins or so)!

Schmavery:
Thanks for helping test it out :)

parkerziegler:
Released in [v1.0.1](https://github.com/FormidableLabs/reason-urql/releases/tag/v1.0.1).

Issue #106, opened 2019-09-27
Update API documentation on CombinedError.


parkerziegler:
Currently our API documentation on `CombinedError` is non-existent. This is an important part of the `urql` API and deserves to be covered in more depth. API documentation should include the following:

- Type signature and description of the `CombinedError.combinedError` record.
- Examples showing how to pattern match `combinedError` on the `networkError` and `graphQLErrors` fields.
- Example showing the `message` field added in #105.

In addition, we should add a `.rei` file for this `module` as long as we're here.

Schmavery:
One thing that I found a little funny with this module is that the "type t" is the js version of this type, which afaict isn't really used that much by users. Mostly I would interact with the `combinedError` type (the record version). Caused some confusion once when trying to annotate something with UrqlCombinedError.t fwiw

parkerziegler:
Definitely agreed, that was an oversight on my part @Schmavery. I think initially I was planning to leave the `Js.t` as the exposed API, but now that the record is what people are actually working with I'm going to adjust the API.

PR #108, opened 2019-10-01
FIx capitalization in CombinedError bindings

Schmavery:
Right now this is always null, since we don't match https://github.com/FormidableLabs/urql/blob/f4b11379e43ef14bbc2a5bf563a6aa2cc09b4271/src/utils/error.ts#L43

Probably we should also update `graphqlErrors` in the record to match, but since that's a breaking change, I thought it might be best done as part of something in #106?

### DIFF
--- a/src/utils/UrqlCombinedError.re
+++ b/src/utils/UrqlCombinedError.re
@@ -53,7 +53,7 @@ class type _combinedError =
   [@bs]
   {
     pub networkError: Js.Nullable.t(Js.Exn.t);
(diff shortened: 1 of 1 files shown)


parkerziegler:
Thanks @Schmavery, good catch! Yep, I'll have some updates for #106 b/t today and tomorrow sometime, so I'm thinking we'll release this plus those changes in a new minor version (since, as you point out, these are all breaking changes). Does that sound good to you?

Schmavery:
Sounds good. Your call if you wanna do breaking changes in a minor version, doesn't affect my usage too much.

PR #109, opened 2019-10-04
Update CombinedError API to expose the combinedError record as the core module type t.

### LINKED ISSUE #106: Update API documentation on CombinedError.

parkerziegler:
Fix #106 

This PR updates our `CombinedError` API to pass the `combinedError` record as the module type `t` that users will interact with. This feels more ergonomic to me and I believe @Schmavery felt the same. I also added the beginnings of docs on how to use `CombinedError`. I decided it wasn't useful to add a `.rei` file for this module b/c it is full of a bunch of BuckleScript accessors generated by all the `[@bs.deriving abstract]`, and we basically want the entirety of the module public anyway.

In terms of how we want to release this, I'm a little torn. It is a small breaking change that will (I suspect) break folks' builds, but it also feels fairly minimal. I'm inclined to release it as a minor (`v1.1.0`) but could also consider a patch (`v1.0.2`) if others feel that works better. cc/ @Schmavery @gugahoa let me know your thoughts here. @Schmavery I co-authored you on the commit by the way, since I slightly modified the error modifier you have on Coda for the example here (it's really nice btw, love it).

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -740,7 +740,16 @@ In this case, `networkError` returns the original JavaScript error thrown if a n
 
 #### Example
 
-Dependening on the types of errors you get from your GraphQL API, you may want to do different things. Here's an example showing how to handle `networkError`s and `graphQLErrors` separately.
+The easiest way to interact with `CombinedError` is through the `message` field on the `CombinedError.t` record. If you just want to display the error message to your user, or operate directly on the error string, simply do the following:
+
+```reason
+switch (response) {
+  | Error(e) => <div> e.message->React.string </div>
+  | _ => <div />
+}
+```
+
+Dependening on the types of errors you get from your GraphQL API, you may want to do different things. Here's an example showing how to handle `networkError`s and `graphQLErrors` indepedently.
 
 ```reason
 let ({ response }, _) = useQuery(~request, ());
@@ -773,15 +782,6 @@ switch (response) {
     | _ => <div> "Unknown error."->React.string </div>
     }
   }
-  | _ => <div />
-}
-```
-
-Due to the verbosity of the above, sometimes it isn't useful to handle every potential case around errors in your application. When you just want a simple error message of what went wrong to display to the user, just access the `message` field on the `CombinedError.t`:
-
-```reason
-switch (response) {
-  | Error(e) => <div> e.message->React.string </div>
   | _ => <div />
 }
 ```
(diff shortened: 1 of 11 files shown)


docs/api.md
```diff
@@ -722,3 +722,66 @@ let client = Client.make(
...
+    | _ => <div> "Unknown error."->React.string </div>
+    }
+  }
+  | _ => <div />
+}
+```
+
+Due to the verbosity of the above, sometimes it isn't useful to handle every potential case around errors in your application. When you just want a simple error message of what went wrong to display to the user, just access the `message` field on the `CombinedError.t`:
+
+```reason
```
Schmavery: For me this ends up being what I use 90% of the time, any thoughts on putting it before the more complex solution?

Schmavery:
The version doesn't make much difference to me personally, I can't imagine not upgrading either way :P

Thanks for adding the docs, it's great to have something for that. Thanks for checking out the coda code :) though ironically writing code like that is replaced easily with e.message (and I actually removed that code from our codebase haha)!
I wonder if it's worth writing something that would return some different kind of react component, or in one case perform some side effect like recording a log message... I guess it doesn't matter much either way.

docs/api.md
gugahoa: It's also how I mainly handle errors in the front-end, I would also prefer if it was above the more complex way to handle errors just for the sake of it and to not confuse beginners

parkerziegler: Sounds good, thanks for the feedback!

parkerziegler:
@Schmavery that's an interesting idea. It would be cool to handle all of the more complex parsing of `graphQLErrors` and `networkError` for users on our end at the very least. Definitely will consider this for a later release ✅

Schmavery:
Sorry, I didn't mean that the library should do that sort of thing, I meant that the example could do something other than stringifying the error, since e.message is already that.

parkerziegler:
I think that'd be a worthwhile example to show. I'm considering starting an `faq.md` for this kind of thing – more difficult interesting / sophisticated examples with common edge cases that people have run into. I think that'd be a great place so that we keep this `api.md` focused on the low-level API.

PR #114, opened 2019-10-20
docs: Add bs-fetch peer dependency to README

cem2ran:
After FormidableLabs#104 was merged, you have to explicitly install bs-fetch which may not be obvious to everyone until you try to compile and it fails:

```bash
File "bsconfig.json", line 1
Error: package bs-fetch not found or built 
- Did you install it?
- If you did, did you run `bsb -make-world`?
error Command failed with exit code 2.
```

So to avoid any friction in installing the library I think it's preferable to make this explicit in step 1 such that you are not met with the error when you try to compile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,10 +24,10 @@ Reason bindings for Formidable's Universal React Query Library, [`urql`](https:/
 
 ## 💾 Installation
 
-#### 1. Install `reason-urql`.
+#### 1. Install `reason-urql` & `bs-fetch` peer dependency.
 
 ```sh
-yarn add reason-urql
+yarn add reason-urql bs-fetch
 ```
 
 #### 2. Add `graphql_ppx`.


parkerziegler:
Hey @cem2ran, thanks for the PR. Interestingly, I don't seem to hit this error when compiling on a new project. Can you try the following steps and let me know if you're still hitting this issue:

1. Run `bsb -init reason-urql-test -theme react-hooks` to create a new ReasonReact project.
2. Add `reason-urql` as a dependency and add it to your `bs-dependencies` array.
3. Paste some minimal code into `index.re`, something like:

```reason
open ReasonUrql;

let client = Client.make(~url="https://localhost:3000", ());

ReactDOMRe.renderToElementWithId(
  <div />,
  "index1",
);
```

The dependency on `bs-fetch` is a bit up in the air for me. It's mostly there for type checking information on the client's `fetchOptions` property, and I would expect to see the error you're reporting. For some reason I'm just not seeing it with a new project. I'll double check as well with the above steps and ensure I can't repro. Thanks again for the PR!

parkerziegler:
@cem2ran, never mind,  you're totally right – the example project I was using to test on had a symlink hanging around. Thanks for the PR, merging now!

parkerziegler:
@all-contributors please add @cem2ran for docs.

PR #116, opened 2019-10-27
Upgrade bs-platform to 5.2.1. Upgrade dependencies to fix security warnings.

parkerziegler:
This PR beings the process of migrating to v1.2.0 of `urql`! v1.2.0 has very little different when it comes to the bindings itself – the devtools features introduced in that release were later moved into the devtools exchange here: https://github.com/FormidableLabs/urql-devtools-exchange. For that reason, I don't see much point in writing those bindings for this release. Instead, I think we'll want to create `bs-urql-devtools-exchange` and use that for future releases. If others feel differently please let me know!

The vast majority of changes in this PR are just upgrading depedencies to fix some security warnings and sprucing up some of the examples both code-wise and UI-wise. I'll do a deeper dive into the 1.2.0 release of `urql` to see if there are any other features we're missing and schedule some time to get this out later this week!

cc/ @Schmavery @gugahoa


parkerziegler:
No worries, thanks for taking a quick peak @gugahoa. I'll cut a release tomorrow! I did end up adding the `operationDebugMeta` information to the `operationContext` on the client, since it is technically a part of `urql` 1.2.0. I think I got everything else of note as well, so we'll test in production and I can cut patch releases if we folks find issues.

PR #125, opened 2019-11-08
Improve API docs, including instructions on graphql_ppx_re.

### LINKED ISSUE #121: New GraphQL ppx instructions incomplete
The way to generate a `graphql_schema.json` is different on `graphql_ppx_re` so the instructions on the README right now will lead to an error when setting up a new project

We should copy the instructions from `graphql_ppx_re` and have another section with instructions if the user decides to use `graphql_ppx` only

parkerziegler:
Fix #121. 

This PR begins the work to improve our API documentation and clarify language around using `graphql_ppx_re`. The biggest hurdle with the new lib is the difference in how a user derives a `graphql_schema.json` file (via `graphql-cli` rather than a local Node script that ships with the lib).
We'll probably want to begin converting our examples over gradually to use the new PPX – perhaps a good issue for a first time contributor.

I also added a bit more to our `getting_started.md` to discuss `useMutation`. Recent changes to the `urql` README added a lot of good information that it may be worthwhile to piggyback off of here as well, so I'll keep working on making the docs more usable. Open to any suggestions on how to improve things.

The last change in this PR is just ensuring we're using `CombinedError.t` instead of `CombinedError.combinedError` to type that record across the lib.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -134,7 +134,7 @@ Great – we've successfully executed a mutation to like a dog! Existing queries
 ```reason
 [@react.component]
 let make = (~key: string) => {
-  /* Build your request by calling .make on the graphql_ppx module, passing variables as labeled arguments. */
+  /* Build your request by calling .make on your mutation, passing variables as labeled arguments. */
   let request = LikeDogMutation.make(~key, ());
 
   /* Pass the request to useMutation. */
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -38,7 +38,7 @@ open ReasonUrql;
 /* Be sure to open the Hooks module to bring necessary types into scope. */
 open Hooks;
 
-/* Create a graphql_ppx module with your GraphQL query. */
+/* Create a module with your GraphQL query. */
 module DogsQuery = [%graphql
   {|
   query dogs {
@@ -54,7 +54,7 @@ module DogsQuery = [%graphql
 
 [@react.component]
 let make = () => {
-  /* Build your request by calling .make on the graphql_ppx module. */
+  /* Build your request by calling .make on your query. */
   let request = DogsQuery.make();
 
   /* Pass the request to useQuery. */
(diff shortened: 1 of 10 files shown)


docs/getting_started.md
```diff
@@ -83,8 +83,87 @@ let make = () => {
...
+  /* Pass the request to useMutation. */
+  let (_, executeMutation) = useMutation(~request, ());
+
+  <button onClick=(_e) => executeMutation()>
+      "Execute the Mutation (and Reward a Good Dog)"->React.string
+  </button>
+}
+```
+
+Great – we've successfully executed a mutation to like a dog! Existing queries that reference the mutated data will be notified that data has changed, meaning you don't need to think at all about refetching your data – it's all updated for you by `reason-urql`.
```
gugahoa: Oh, this is quite nice. I didn't know that it worked like this :thinking: 
How do I make sure I'm using this feature?

docs/getting_started.md
```diff
@@ -83,8 +83,87 @@ let make = () => {
...
+
+`useMutation` returns a two-dimensional tuple, containing `(result, executeMutation)`. `result` contains the `response` variant, which allows you to pattern match on the API response from your mutation. For example, if you wanted to show different UI when your mutation was `Fetching`, or if there was an `Error(e)` you can do something like the following:
+
+```reason
+[@react.component]
+let make = (~key: string) => {
+  /* Build your request by calling .make on the graphql_ppx module, passing variables as labeled arguments. */
```
gugahoa: Here you mention `graphql_ppx` but in other places you removed the mention

docs/getting_started.md
parkerziegler: It has to do with `urql`'s observable system, which is powered by [`wonka`](https://github.com/kitten/wonka). Essentially, every operation that passes through the Client can have subscribers, and this is particularly important in the case of queries that are dependent on mutation results. When a mutation is executed, `urql` will aggressively invalidate the cache for any entity that has a shared `__typename` with the mutation. This then notifies subscribers to those cache entries (really any query that requests data with the same `__typename`) that they should update: https://github.com/FormidableLabs/urql#document-caching. The first example project in this repo shows this pretty well – if you hit the button to fire `executeMutation`, the query will be notified that the cache was invalidated and pull anew from it, causing a re-render. Hope that makes sense! You don't need to do anything to opt into this, it should just happen for you.

docs/getting_started.md
parkerziegler: Good catch, fixed here: f67eb9f5216b3f82dc4f893edfb024ccc15bffe8

PR #127, opened 2019-11-10
Migrate to bs-platform 6.2.1. Fix compiler warnings. Migrate examples.

### LINKED ISSUE #126: Upgrade to bs-platform 6.2.1.
With `bs-platform@6.2.1` out, we'll want to start migrating our in app examples over to this version to ensure compatibility. This will also entail upgrading the root's `bs-platform` `devDependency` to 6.2.1. Finally, this also means that all examples will need to use `@baransu/graphql_ppx_re/ppx6`.

parkerziegler:
Fix #126 

This PR migrates our `devDependency` on `bs-platform` to 6.2.1, which ensures that the next release of `reason-urql` will be fully compatible with the latest BuckleScript compiler. @gugahoa got the major bug in #124, so the changes here are mostly just to fix warnings and add additional code comments for contributors. If this looks good to folks I'll merge and publish v1.2.1.

I also took the time to upgrade our examples to `bs-platform` 6.2.1 and switched our PPX for GraphQL over to `@baransu/graphql_ppx_re/ppx6`, which appears to work quite well.

### DIFF
--- a/src/UrqlClient.re
+++ b/src/UrqlClient.re
@@ -252,19 +252,19 @@ let make =
 
 /* A function to convert the JS response from the client to a Reason record. */
 let urqlClientResponseToReason =
-    (~parse: Js.Json.t => 'response, ~result: ClientTypes.operationResult) => {
-  open ClientTypes;
-
-  let data = result->dataGet->Js.Nullable.toOption->Belt.Option.map(parse);
+    (~parse: Js.Json.t => 'response, ~result: ClientTypes.operationResult)
+    : ClientTypes.clientResponse('response) => {
+  let data =
+    result->ClientTypes.dataGet->Js.Nullable.toOption->Belt.Option.map(parse);
   let error =
     result
-    ->errorGet
+    ->ClientTypes.errorGet
     ->Js.Nullable.toOption
     ->Belt.Option.map(UrqlCombinedError.combinedErrorToRecord);
 
   let response =
     switch (data, error) {
-    | (Some(data), _) => Data(data)
+    | (Some(data), _) => ClientTypes.Data(data)
     | (None, Some(error)) => Error(error)
     | (None, None) => NotFound
     };
--- a/src/UrqlTypes.re
+++ b/src/UrqlTypes.re
@@ -41,10 +41,10 @@ type hookResponse('ret) = {
 };
 
 [@bs.deriving abstract]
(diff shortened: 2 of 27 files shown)


src/UrqlClient.re
```diff
@@ -252,19 +252,19 @@ let make =
...
     result
-    ->errorGet
+    ->ClientTypes.errorGet
     ->Js.Nullable.toOption
     ->Belt.Option.map(UrqlCombinedError.combinedErrorToRecord);
 
   let response =
     switch (data, error) {
-    | (Some(data), _) => Data(data)
+    | (Some(data), _) => ClientTypes.Data(data)
```
parkerziegler: The majority of warnings we were getting with the upgrade were related to shadowed variant constructors due to our use of `open`. It seems like slightly better practice to avoid global `open` here to ensure the proper variant constructors get picked up – but definitely open to discussion here!

src/UrqlTypes.re
```diff
@@ -41,10 +41,10 @@ type hookResponse('ret) = {
 };
 
 [@bs.deriving abstract]
-type jsResponse = {
+type jsResponse('response) = {
```
parkerziegler: Passing a type argument here allows us to use this type across all hooks – previously `useSubscription` couldn't use this type because the JS response from calling the `useSubscription` hook could be either a `Js.Json.t` (if no handler was provided), or the return type of the handler function (if a handler was provided). This just makes it more flexible.

src/hooks/UrqlUseQuery.rei
```diff
@@ -1,22 +1,15 @@
-open UrqlTypes;
-
-type partialOperationContextFn =
+type executeQuery =
```
gugahoa: Isn't this a breaking change? Since it was exposed to the public, someone could be using it and the next release would inadvertently break their code

parkerziegler: I suppose that's true. They'd need to be referencing the `type` directly, but like you said that's technically part of the public API. I think we should probably release this as 1.3.0 anyway with a notice about `bs-platform@6.2.1` support. It is an intended change as I think it gives better context around what that function actually is (the `executeQuery` function with support for altering the query execution environment).

gugahoa: Ok, sounds good to me!

Issue #128, opened 2019-11-11
Passing variables only when calling executeMutation


gugahoa:
In the PR https://github.com/FormidableLabs/reason-urql/pull/69 by @Schmavery where typed use mutation was introduced, there was also a suggestion on making the interface for `useMutation` take `makeWithVariables` as parameter to let users pass variables later. In hindsight, it seems this is the expected behavior by users as can be seen here: https://spectrum.chat/urql/reason/usemutation-api~72b6306a-8d92-4d17-9fff-15826dd64070 and here https://github.com/FormidableLabs/reason-urql/issues/103

What are the thoughts on creating another `useMutation`/`useQuery` function where it's possible to pass the variables when calling `executeMutation`/`executeQuery`?

cc @parkerziegler

parkerziegler:
Alright, so diving back into this, I feel like the two implementations I'm seeing discussed fall into two categories.

1. Use different parts of the `graphql_ppx` / `graphql_ppx_re` API to achieve what we want. This is the implementation Avery mentioned in #69, which I think is a cool idea. Rather than passing `.make()` with variables applied, we pass the `makeWithVariables` function itself.

```reason
let (result, executeMutation) = Hooks.useMutation(~request=MyQuery.makeWithVariables, ());
```

In this case `executeMutation` would essentially use `makeWithVariables` under the hood. What seems a bit difficult with this implementation is that I don't yet see a way to grab `MyQuery.t` off `makeWithVariables` without actually applying it, which means typing `result` would be difficult. This would also mean that users become responsible for parsing the returned `result`. Right now in `useMutation` we get access to that `parse` function by applying `make`, and we use that to `parse` the `response` for the user before handing it back to them.

2. We go for the functor approach similar to `reason-apollo-hooks`. @gugahoa scaffolded out this API a bit on Spectrum:

```reason
module type QuerySig {
  type t;
  let parse: Js.Json.t => t;
  let query: string;
};

let useMutation:
  (~request: (module QuerySig with type t = 'response)) =>
  (
    hookResponse('response),
    Js.Json.t => Js.Promise.t(UrqlClient.ClientTypes.operationResult)
  )
```

[...]

PR #129, opened 2019-11-19
Add functor API for hooks.

parkerziegler:
This PR adds a functor API for `reason-urql`'s hooks. The goal of the functor API is to address concerns raised in #128 concerning dynamic variables for mutations, where we want to pass variables at execution time rather than when the hook is ran. This is closer to how `urql` actually works – the `executeMutation` function returned by the `useMutation` hook accepts an optional `variables` argument that will be applied when the function is called.

The functor API would look to consumers like the following:

```reason
module LikeDog = [%graphql {|
    mutation likeDog($id: ID!) {
       likeDog(id: $id) {
           id
           name
           likes
       }
    }
|}];

[@react.component]
let make = (~id: string) => {
   module LikeDogMutation = ReasonUrql.Hooks.MakeMutation(LikeDog);
   let (result, executeMutation) = LikeDogMutation.useMutation();

   let payload = Js.Dict.empty();
   Js.Dict.set(payload, "id", Js.Json.string(id));

  <button onClick={_ev => executeMutation(Some(Js.Json.object_(payload))) |> ignore}>
     "Execute Mutation"->React.string
  </button>
};
```

In addition to adding a functor API for `useMutation`, I also added them for `useQuery` and `useSubscription`. This is more to offer folks the option of using the functor APIs if they like them, since folks coming from `reason-apollo` or `reason-apollo-hooks` will likely be most used to that approach.

Ideally, I'd love for us to find a way of achieving the above without the use of functors to keep the API simpler, but at this moment I think the functor API gets us closest to how `urql` actually works without sacrificing type inference. I'm very curious to see what people think here.

**Note:** Don't be alarmed by the "breaking build" in these commits. GitHub changed its billing structure and Formidable's plan doesn't give us access to GitHub Actions. So we're going to be switching back to Travis here soon.

cc/ @robinweser tagging you as well to take a peek!

### DIFF
--- a/src/hooks/UrqlUseMutation.re
+++ b/src/hooks/UrqlUseMutation.re
@@ -61,3 +61,46 @@ let useMutation = (~request) => {
 
   (response, executeMutation);
 };
+
+module type MutationConfig = {
+  type t;
+  let parse: Js.Json.t => t;
+  let query: string;
+};
+
+module type MakeType =
+  (Mutation: MutationConfig) =>
+   {
+    let useMutation:
+      unit =>
+      (
+        UrqlTypes.hookResponse(Mutation.t),
+        React.callback(
+          Js.Json.t,
+          Js.Promise.t(UrqlClient.ClientTypes.operationResult),
+        ),
+      );
+  };
+
+module Make: MakeType =
+  (Mutation: MutationConfig) => {
+    let useMutation = () => {
+      let (responseJs, executeMutationJs) = useMutationJs(Mutation.query);
+
+      let response =
+        React.useMemo2(
+          () => responseJs |> urqlResponseToReason(Mutation.parse),
+          (Mutation.parse, responseJs),
+        );
+
+      let executeMutation =
+        React.useCallback1(
+          (variables: Js.Json.t) => executeMutationJs(Some(variables)),
+          [|executeMutationJs|],
+        );
+
+      (response, executeMutation);
+    };
+
+    useMutation;
+  };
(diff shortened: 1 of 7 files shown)


src/hooks/UrqlUseMutation.re
```diff
@@ -54,10 +54,63 @@ let useMutation = (~request) => {
     );
 
   let executeMutation =
-    React.useCallback1(
+    React.useCallback2(
       () => executeMutationJs(Some(request##variables)),
-      [|request##variables|],
+      (executeMutationJs, request##variables),
```
parkerziegler: It seemed like this should be in the dependency array, since it's a function returned from a hook and its identity is not guaranteed.

src/hooks/UrqlUseSubscription.re
```diff
@@ -95,3 +95,63 @@ let useSubscription =
...
+  type resp;
+
+  let useSubscription:
+    (~variables: Js.Json.t=?, ~handler: handler('acc, resp, 'ret)) =>
+    UrqlTypes.hookResponse('ret);
+};
+
+module MakeSubscription =
+       (Subscription: SubscriptionConfig)
+       : (MakeSubscriptionType with type resp = Subscription.t) => {
```
parkerziegler: This one through me for a bit of a loop. I had to learn a bit about _sharing contraints_ in OCaml – see more in the **Destructive Substitution** section here: https://v1.realworldocaml.org/v1/en/html/functors.html Basically, we need to be able to pass the type `t` of the `module` passed _into_ the `MakeSubscription` functor to the `module type MakeSubscriptionType` such that the `resp` type is inferred properly. Quite fun!

parkerziegler:
Closing in favor of #130.

PR #130, opened 2019-11-21
Add useDynamicMutation hook

Schmavery:
Wrote up some code for the useDynamicMutation idea in #128
Updated the examples to show usage but had some trouble running them (webpack acting up) so it would be great if someone can double-check that it works, otherwise I'll figure it out some other time.

### DIFF
--- a/examples/3-mutation/src/Dog.re
+++ b/examples/3-mutation/src/Dog.re
@@ -62,9 +62,11 @@ let make =
       [|id|],
     );
 
+  // Example of using hooks with graphql_ppx
   let (_, executeLikeMutation) =
     Hooks.useMutation(~request=Mutations.LikeDog.make(~key=id, ()));
 
+  // Example of using hooks without graphql_ppx
   let (_, executeTreatMutation) =
     Hooks.useMutation(
       ~request={
@@ -72,6 +74,14 @@ let make =
         "variables": payload,
         "parse": x => x,
       },
+    );
+
+  // Example of using hooks where the variables are only known when the
+  // mutation runs
+  let (_, executePatMutation) =
+    Hooks.useDynamicMutation(
+      ~query=Mutations.PatDog.query,
+      ~parse=Mutations.PatDog.parse,
     );
 
   <div className=DogStyles.container>
@@ -84,22 +94,21 @@ let make =
         hex="48a9dc"
         onClick={_ => executeLikeMutation() |> ignore}
       />
-      <Mutation request={Mutations.PatDog.make(~key=id, ())}>
-        ...{({executeMutation}) =>
-          <EmojiButton
-            emoji={j|✋|j}
-            count={string_of_int(pats)}
-            hex="db4d3f"
-            onClick={_ => executeMutation() |> ignore}
-          />
+      <EmojiButton
+        emoji={j|✋|j}
+        count={string_of_int(pats)}
+        hex="db4d3f"
+        onClick={_ =>
+          executePatMutation(Mutations.PatDog.make(~key=id, ())) |> ignore
         }
-      </Mutation>
+      />
       <EmojiButton
         emoji={j|🍖|j}
         count={string_of_int(treats)}
         hex="7b16ff"
         onClick={_ => executeTreatMutation() |> ignore}
       />
+      // Example of using the Mutation component
       <Mutation request={Mutations.BellyscratchDog.make(~key=id, ())}>
         ...{({executeMutation}) =>
           <EmojiButton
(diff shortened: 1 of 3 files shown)


src/hooks/UrqlUseMutation.re
```diff
@@ -61,3 +61,21 @@ let useMutation = (~request) => {
...
+
+  let response =
+    React.useMemo1(
+      () => responseJs |> urqlResponseToReason(parse),
+      [|responseJs|],
+    );
+
+  let executeMutation =
+    React.useCallback1(
+      varsObj => executeMutationJs(Some(varsObj##variables)),
```
parkerziegler: A nit on naming here. This is the `Js.t` created by calling `.make()` on the `graphql_ppx_re` module, yeah? Could we use the same nomenclature we do in `useMutation` and name this `request`?

```suggestion
      request => executeMutationJs(Some(request##variables)),
```

src/hooks/UrqlUseMutation.rei
```diff
@@ -4,3 +4,11 @@ let useMutation:
     UrqlTypes.hookResponse('response),
     unit => Js.Promise.t(UrqlClient.ClientTypes.operationResult),
   );
+
+let useDynamicMutation:
+  (~query: string, ~parse: Js.Json.t => 'response) =>
+  (
+    UrqlTypes.hookResponse('response),
+    {.. "variables": Js.Json.t} =>
```
parkerziegler: Is there any harm in defining this type as:

```suggestion
    UrqlTypes.request('response) =>
```

That would match the signature of calling `.make` exactly, even tho we never access those other two fields. Or is the idea to explicitly _not_ include those fields so they can't be accessed?

examples/3-mutation/src/Dog.re
```diff
@@ -62,9 +62,11 @@ let make =
       [|id|],
     );
 
+  // Example of using hooks with graphql_ppx
```
parkerziegler: I think we should mention both PPXs in case someone stumbles across this and thinks the lib only works with `graphql_ppx`.

```suggestion
  /* Example of using hooks with graphql_ppx / graphql_ppx_re. */
```

parkerziegler:
Also, I applied the patch locally and it works great. Yeah, `webpack` acts up a lot in those examples. I'm thinking I should move to a `yarn workspaces` setup so we can get rid of the dependency duplication and have more reliable local dev.

src/hooks/UrqlUseMutation.re
Schmavery: The type is slightly different (explained below), which is why I went with a different name. Doesn't bother me to change it though, if you think it's clearer.

src/hooks/UrqlUseMutation.rei
Schmavery: I thought that since we only use the variables, we don't need to require that users provide the other fields. This way, if someone can't use graphql_ppx for some reason, they can do this:

```reason
executeMutation({"variables": {"var1": "hello", "var2": "world"}});
```
instead of
```reason
executeMutation({
  "variables": {"var1": "hello", "var2": "world"},
  "query": "",
  "parse": a => a
});
```

examples/3-mutation/src/Dog.re
Schmavery: Do you prefer non-single-line comments?

parkerziegler: I think I just tend to use `/* */` mostly out of habit from the Reason docs, I actually didn't know that single line comments were supported. Happy either way!

src/hooks/UrqlUseMutation.re
parkerziegler: Ah that makes sense. I really wish we could destructure "open" objects like this so it'd be more like:

```reason
{ variables } => executeMutationJs(Some(variables))
```

Still I do find `request` a little easier to understand. This object passed to this callback is the result of calling `.make()` on that `graphql_ppx_re` module, and we have been calling that `request` throughout the codebase. Even tho we're restricting its type a bit more, it's more or less the same thing.

src/hooks/UrqlUseMutation.rei
parkerziegler: Ah right you are, good shout!

Schmavery:
Updated the variable name and comments, will merge tomorrow morning if there are no more concerns by then (or feel free to beat me to it).

robinweser:
This looks brilliant! Happy to test it out :)

parkerziegler:
Looks great to me @Schmavery. I'll write up some docs for it this afternoon and get this released today or tomorrow! Thanks so much!

PR #131, opened 2019-11-24
Restore Travis CI.

parkerziegler:
Since Formidable has a legacy GitHub organizational account, we save quite a bit of money compared to newer GitHub accounts. However, with our plan we don't have access to GitHub Actions 😭 A shame since they were working so nicely in our repo, but not a big deal since Travis should work just fine. This PR restores Travis as our CI provider.

The steps are the same as the GitHub Actions workflow. We build and test, in addition to building each of the examples to ensure they still compile with latest changes.


ryan-roemer:
Lgtm. Did something go wrong with the other node versions?

parkerziegler:
@ryan-roemer No, but when I had multiple Node versions in there the build would fail because it would only run `yarn build` for one version (typically v8). For the others it would immediately run `yarn test`, which won't work b/c the `build` command is what actually runs the BuckleScript compiler and generates the JS files that `jest` will run against. Having broad matrix support doesn't feel critical to me – LTS is fine.

Here's an example run: https://travis-ci.org/FormidableLabs/reason-urql/jobs/616401238?utm_medium=notification&utm_source=github_status

If you look at the config file, you'll see that the Test step includes:

```yml
- yarn build
- yarn test
```

But in the log itself I only see the `yarn install` run and then it immediately runs `yarn test` (and thus can't find any tests b/c they haven't been compiled).

PR #132, opened 2019-11-25
Improved useDynamicMutation API

Schmavery:
Jumping the gun and posting this WIP PR that @sgrove wrote with me today.
This is dependent on a proposed change to graphql_ppx_re (https://github.com/baransu/graphql_ppx_re/pull/48) that would add a `definition` variable containing the query, parse function and a special `composeVariables` function that lets us do cool API things.

Wanted to get some early feedback on the first one (useDynamicMutation) that we converted, but I'm also working on similar changes to useQuery/useMutation etc which could be neat. It seems to work well so far but I'm very interested if people spot any potential issues.

In case anyone is interested, though it looks a little complicated, this shouldn't make it any harder for non-graphql_ppx_re users to use the library. They would have to write something like this:

```reason
useQuery((a => a, "query", fn => fn(Js.Json.null)));

// We could even expose a helper function like this:
let jsonVars = (v, f) => f(v);
useQuery((a => a, "query", jsonVars(Js.Json.null)));

// Or like this..
let makeDefinition = (~parse: Js.Json.t => args, ~query: string) =>
  (parse, query, (f, vars: Js.Json.t) => f(vars));
Hooks.useQuery(Hooks.makeDefinition(~parse=a=>a, ~query="query"), Js.Json.null);
```

PR #133, opened 2019-11-25
Port new hooks API to other hooks

Schmavery:
This depends on https://github.com/FormidableLabs/reason-urql/pull/132! Read that one first.

This change still needs me to update the examples (I'm waiting until https://github.com/baransu/graphql_ppx_re/pull/48 lands), but here's the usage:
```reason
let (response, refetch) = Hooks.useQuery(MyQuery.definition, ~var=1, ());
let (response, refetch) = Hooks.useMutation(MyMutation.definition, ~var=1, ());
let (response, execute) = Hooks.useDynamicMutation(MyMutation.definition);
let (response) = Hooks.useSubscription(MySub.definition, ~var=1, ());
```

Curious about people's opinions on how to do a breaking change like this. Personally I think I'd be ok with a breaking version change at some point, but we could also copy these into another Hooks module or something if people prefer I guess.


src/hooks/UrqlUseMutation.re
```diff
@@ -72,8 +77,8 @@ let useDynamicMutation = (~query, ~parse) => {
     );
 
   let executeMutation =
-    React.useCallback1(
-      request => executeMutationJs(Some(request##variables)),
+    React.useMemo1(
+      () => composeVariables(request => executeMutationJs(Some(request))),
       [|executeMutationJs|],
```
gugahoa: I noticed in the other PR, but `composeVariables` is missing from the dependency list of `useMemo`

Schmavery: oh fair enough, thanks

jfrolich:
Hi, I've been playing with this API for `reason-apollo-hooks`. It looks like the currying works well, but there is a compiler error when variables are not provided in the same order as in the query/mutation. At least in my test implementation which is almost the same as this one. I am not really sure why that is, but maybe worth looking into (and finding a fix if possible).

src/hooks/UrqlUseMutation.re
parkerziegler: This is also somewhat of an interesting question. Functions should go in the dependency array if they close over any piece of data that participates in the React lifecycle. In this case, `composeVariables` is passed into the hook by that `definition` argument, which is provided by the user. `composeVariables` then closes over the value of `executeMutationJs`, which I think is why we need it in the dep. array (in addition, the linting rules for hooks will tend to treat _any_ function passed into a custom hook as needing to be in the dep . array in my experience). But good to know that if we were referencing a utility function of some kind from some other module, it wouldn't necessarily need to be included in this dep. array.

src/hooks/UrqlUseMutation.re
```diff
@@ -41,28 +41,33 @@ let urqlResponseToReason =
  *
  * Accepts the following arguments:
  *
- * request – a Js.t containing the query and variables corresponding
- * to the GraphQL mutation, and a parse function for decoding the JSON response.
+ * request – a tuple containing the query, a parse function for decoding the
+ * JSON response, and a higher-order function that calls a passed-in function
+ * with the variables corresponding to the GraphQL. This should be generated by
```
parkerziegler:
```suggestion
 * with the variables corresponding to the GraphQL mutation. This should be generated by
```

src/hooks/UrqlUseMutation.re
```diff
@@ -41,28 +41,33 @@ let urqlResponseToReason =
...
-  let executeMutation =
-    React.useCallback1(
-      () => executeMutationJs(Some(request##variables)),
-      [|request##variables|],
-    );
+    let executeMutation =
+      React.useCallback1(
+        () => executeMutationJs(Some(variables)),
+        [|variables|],
+      );
```
parkerziegler: A question here that I'm sure we solved at some point but for some reason is slipping my mind – what about mutations that don't accept any variables? It seems like this impl. (and our old one too) assume that you have `variables`. Does `graphql_ppx_re` reliably return no `variables` as something like `Js.Json.object_(Js.Dict.empty())`? Just wanting to make sure we handle the no variable mutation case which, while rare, does exist.

src/hooks/UrqlUseMutation.rei
```diff
@@ -1,14 +1,18 @@
...
-    UrqlTypes.hookResponse('response),
-    unit => Js.Promise.t(UrqlClient.ClientTypes.operationResult),
-  );
+  UrqlTypes.graphqlDefinition(
+    'parsedResponse,
+    (
+      UrqlTypes.hookResponse('parsedResponse),
+      unit => Js.Promise.t(UrqlClient.ClientTypes.operationResult),
+    ),
+    'functionThatAcceptsVarArgsAndReturnsHookResponse,
```
parkerziegler: Can we find a more concise name for this type variable? Would `'composeVariables` be appropriate here?

Schmavery: it's not really the composeVariables function though. I tried to come up with a name that clearly defines what it is but couldn't come up with anything shorter than this.
I could call also call it `'a`? haha

parkerziegler: Ah ok, I guess I'm misinterpreting how `composeVariables` works. So the idea is that `composeVariables` returns you a **function**, which accepts your `variables` as a `Js.Json.t`. I guess I'm missing the step around how variables that we pass in as labeled arguments get passed to `composeVariables` as a `Js.Json.t`:

```reason
Hooks.useMutation(MyMutation.definition, /* How are we able to pass these in and get a Js.Json.t? */ ~var1=x, ~var2=y, ());
```

Is the PPX doing this magic for us? It looks like it from your description in #132. If that's the case, yeah I agree it's difficult to find a real name for this. I'll keep thinking this one over, just wanted to make sure I understand what this is doing.

Schmavery: composeVariables (the thing provided by the ppx), is a higher-order function that accepts a function from Js.Json.t => 'a and returns a new function that accepts all the graphql vars as named args and then calls your passed-in function with a Js.Json.t and returns the result.

We use this by wrapping our hook internals in a function that accepts variables, and then passing this to composeVarialbes and returning the result. This makes useMutation a function that accepts one argument and returns a function of `(~var1, ~var2... ()) => hookResponse`. Because of the magic of auto-currying etc, the user doesn't see this and it feels like they're just calling one function called useMutation.

---

Another way to explain it is to say that the simplest composeVariables function would be something like this:
```
let composeVariables = (f, v) => f(v);
Hooks.useMutation((a => a, "query", composeVariables), Js.Json.null); 
```

useMutation partially applies the composeVariables we pass in with the function that should be called with the variables object, then returns the a function that still needs to be passed `v`. We then immediately do so (`Js.Json.null`) in our call to "useMutation".
Passing `Js.Json.null` fully applies the `composeVariables` we passed in, causing useMutation's callback to get called, which returns the hookResponse.

Schmavery:
I tried to leave a better explanation of composeVariables in a comment and hope to use some of this experience attempting to explain it to write some clear docs to add to this PR.

Definitely seems like if we're going to swap this API over, it would make sense to have it be a breaking change. Moving useDynamicMutation -> useMutation is fine with me, I can probably do that here too, if you want.

To be clear, this code is a little rough still, I put up the PR to get some feedback and let people have a look, but I haven't had a chance to test it much yet. I can put a "WIP" in the title if people prefer. That said, I hope to have the time to iron out the wrinkles sometime this week.

jfrolich:
> Hi, I've been playing with this API for `reason-apollo-hooks`. It looks like the currying works well, but there is a compiler error when variables are not provided in the same order as in the query/mutation. At least in my test implementation which is almost the same as this one. I am not really sure why that is, but maybe worth looking into (and finding a fix if possible).

See this commit for a reproduction of this issue in this branch: https://github.com/jfrolich/reason-urql/commit/d984c39ad42244c74f0f3239279fcc4c3b431b03.

This means that with this new API you lose the flexibility of providing the variables in any order, and instead it needs to be exactly as in the query definition. Perhaps no deal breaker, and perhaps there are ways to get around this issue with the OCaml provider.

Schmavery:
@jfrolich that really sucks about the labelled argument order. Does your first-class-module-based solution help with that part somehow or is it the same deal?

What's funny is that it does work when you store the partially applied value to a variable and then call it again like
```
let f = useQuery(GetPokemon.definition);
let ({response}, _) = f(~id="1", ~name=pokemon, ());
```
Not exactly a solution to the problem though. Will have to give this some more thought.

jfrolich:
> @jfrolich that really sucks about the labelled argument order. Does your first-class-module-based solution help with that part somehow or is it the same deal?
> 
> What's funny is that it does work when you store the partially applied value to a variable and then call it again like
> 
> ```
> let f = useQuery(GetPokemon.definition);
> let ({response}, _) = f(~id="1", ~name=pokemon, ());
> ```
> 
> Not exactly a solution to the problem though. Will have to give this some more thought.

That is interesting. The first class modules won't help, that is just an alternative to passing in `Query.definition`. My experience with the reason compiler is limited, but perhaps we can do some extra type annotation to make this work. Would be great to get someone with more experience to look at this (I did some trial and error and couldn't get it to work).

If the currying approach is definitely not possible due to compiler limitations I think the only solution is to pass in a variables labeled argument and use `Query.makeVariables` to provide the variables.

jfrolich:
It might have to do with the type signature of the composeVariables function. Perhaps we need to explicitly generate it as part of `graphql_ppx_re`. 

See:
image

src/hooks/UrqlUseMutation.rei
parkerziegler: Thanks for the detailed explanation – I always forget about auto-currying in Reason spending so much of my day in JS land. That all makes perfect sense 🎉 A very clever implementation and a great usage of the benefits of an FP language!

parkerziegler:
@Schmavery thanks for the clarification on the purpose of the PR // current state of things and @jfrolich for chiming in on the issues with variable ordering. Definitely will read up more on that conversation in Discord. It sounds like potentially enforcing use of `makeWithVariables` in lieu of labeled arguments could help to alleviate the current issue – otherwise the docs would have to be super clear that variables **must be provided in order**, which I agree is not a great restraint to put on API users.

BTW I'll be offline for most of the next four days, so @Schmavery feel free to keep pushing commits here and I'll check in here and there. Next week we can sync on a path forward. Overall I like this API a lot and see it as a major improvement from what we currently have. A 2.0.0 release with this alone would be beneficial, and we can release patch and minor versions as we continue to uncover more about `composeVariables` on the `graphql_ppx_re` side.

parkerziegler:
👋 @Schmavery just wanted to check in on the status of this PR. I definitely want to get this in, I've just been swamped for the past few days and around the holiday. What all do you think we need to get this to a point of being merge-able. I think we're pretty close, and we can pursue docs updates and examples updates in a separate PR to keep the diffs easier to review.

Schmavery:
@parkerziegler sorry, I've been pretty swamped too. I'm a little at a loss about what to do about this because of the discovery of needing to apply the named args in order, which feels worse than what we had in a way. The nice thing is that all the functions become very similar, but I don't know if the tradeoff is worth it.

Ironically we could use this idea to replicate the functor api in non-functor-land but I'm not sure that that's useful at all haha.
```
let `GraphQL(useQuery) = makeQuery(GetPokemon.definition);
let ({response}, _) = useQuery(~id="1", ~name=pokemon, ());
```

---

I wonder if it makes more sense to use makeVariables like reason-apollo-hooks did and deprecate/remove the composeVariables experiment from graphql_ppx_re.
```
let ({response}, _) = useQuery(GetPokemon.definition, GetPokemon.makeVariables());
```

I half-wish that makeVariables returned the Js.t object instead of the Js.Json.t so that for simple queries without optional args we could do:
```
let ({response}, _) = useQuery(GetPokemon.definition, {"id": 123});
```

Sorry, I know that this speculation doesn't help things get merged haha.

parkerziegler:
No that's helpful feedback, thanks for addressing it. I'll play around with the `makeVariables` API a la `reason-apollo-hooks`, since that seems like the best option. In a lot of ways, it also more closely mimics the `urql` API, which I see as a core benefit of that approach. The main goal with next release is really just to provide some avenue for users to be able to supply variables at mutation execution rather than mutation definition. If we can achieve that with `makeVariables` then I'm happy (even tho this really was a superb API and I wish we could ship it – maybe one day we will). Are you alright with closing this PR for now?

parkerziegler:
Also, FYI I tested that `useDyanmicMutation` currently on `master` _does_ work with multiple variables and with providing those variables in different order. I'm inclined to release that API as 1.4.0, since its backwards compatible and solves the immediate use case of needing to apply variables at execution time. I'll put up a PR with the last few changes need there (i.e. docs and the like).

Schmavery:
Yeah sounds ok, I can close this for now.
Makes sense that useDynamic works with different order, seems fine to ship. Thanks for getting this stuff over the finish line

PR #136, opened 2019-12-08
Prepare v1.4.0 with support for useDynamicMutation.

parkerziegler:
This PR simply updates our examples to use the latest version of `graphql_ppx_re` and includes a missing dependency in `useMemo` inside of `useDynamicMutation`. Additionally, it updates the docs to make usage of `useDynamicMutation` clearer. Altogether, I think this is enough to launch v1.4.0. `useDynamicMutation` is a new addition that users can take advantage of, but the other hooks have stayed intact, hence the minor release.

My goal here is to support @robinweser's use case first. I think this will unblock a lot of `reason-urql` users who are  dependent on being able to supply variables at execution time. While I agree a larger hooks refactor to consolidate these APIs may be in order, I don't want to hold up this core feature while we refine that. I'm also optimistic that we'll one day be able to come back to the API @Schmavery suggested in #133, but I want to make sure we get this in ASAP.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Reason bindings for Formidable's Universal React Query Library, [`urql`](https:/
 - ⚛️ A fully featured GraphQL client for ReasonReact.
 - ✅ Compile time type and schema validation.
 - ⚙️ Customizable behavior via `exchanges`.
-- 🎣 Support for `useQuery`, `useMutation`, and `useSubscription` hooks!
+- 🎣 Support for `useQuery`, `useMutation`, `useDynamicMutation`, and `useSubscription` hooks!
 
 `reason-urql` is a GraphQL client for ReasonReact, allowing you to hook up your components to queries, mutations, and subscriptions. It provides bindings to `urql` that allow you to use the API in Reason, with the benefits of a sound type system, blazing fast compilation, and opportunities for guided customization.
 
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -127,7 +122,7 @@ Since we are `link`ing the examples' dependency on `reason-urql` to the `src` di
 
 ## Getting Involved
 
-This project is currently under active development. Please help out by [opening an issue](https://github.com/FormidableLabs/reason-urql/issues) or [filing a PR](https://github.com/FormidableLabs/reason-urql/pulls).
+Please help out by [opening an issue](https://github.com/FormidableLabs/reason-urql/issues) or [filing a PR](https://github.com/FormidableLabs/reason-urql/pulls).
```
parkerziegler: This project still is under active development, but we at Formidable are sticking to using the Maintenance Status section as our source of truth for where maintenance is at and, since this is still experimental, I decided to remove this sentence here (but don't worry, I'm still making changes and continuing to bind new versions of `urql`).

docs/api.md
```diff
@@ -169,7 +169,7 @@ module LikeDog = [%graphql
 [@react.component]
 let make = () => {
   let ({ response }, _) = useQuery(~request=GetAllDogs.make(), ());
-  let (_, executeMutation) = useMutation(~query=LikeDog.query, ~parse=LikeDog.parse);
```
parkerziegler: This was just left over from, I think, a past API proposal. But it's quite incorrect, so I made sure to correct it while here 😅

examples/1-execute-query-mutation/package.json
```diff
@@ -17,7 +17,7 @@
     "reason-urql": "link:../../"
   },
   "devDependencies": {
-    "@baransu/graphql_ppx_re": "^0.3.2",
+    "@baransu/graphql_ppx_re": "^0.4.6",
```
parkerziegler: Picking up latest in all examples so we can get access to `module.definition`.

PR #138, opened 2019-12-15
fix(useQuery): requestPolicy from Reason type to string when passing to urql

### LINKED ISSUE #137: Invalid requestPolicy passed to useQueryJs
```reason
[@bs.deriving abstract]
type useQueryArgs = {
  query: string,
  variables: Js.Json.t,
  [@bs.optional]
  requestPolicy: UrqlTypes.requestPolicy,
  [@bs.optional]
  pause: bool,
};
```
This type doesn't convert nested `UrqlTypes.requestPolicy` type when creating. 

When doing:
```reason
let args =
    useQueryArgs(
      ~query=request##query,
      ~variables=request##variables,
      ~requestPolicy?,
      ~pause?,
      (),
    );
```
is should be 
```reason
let args =
    useQueryArgs(
      ~query=request##query,
      ~variables=request##variables,
      ~requestPolicy=requestPolicy->Belt.Option.map(UrqlTypes.requestPolicyToJs),
      ~pause?,
      (),
    );
```
This way string instead of int will be passed int `"requestPolicy"` key in `urql` options.

baransu:
Fixes #137

### DIFF
--- a/src/hooks/UrqlUseQuery.re
+++ b/src/hooks/UrqlUseQuery.re
@@ -4,7 +4,7 @@ type useQueryArgs = {
   query: string,
   variables: Js.Json.t,
   [@bs.optional]
-  requestPolicy: UrqlTypes.requestPolicy,
+  requestPolicy: string,
   [@bs.optional]
   pause: bool,
 };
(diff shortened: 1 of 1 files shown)


parkerziegler:
@all-contributors please add @baransu for code, bug.

PR #143, opened 2020-01-31
add unit to `examples/5-subscription/src/app/Circles.re`

idkjs:
Subscription example needs `unit` call to handle newly add `Client.ClientTypes.partialOperationContext`


parkerziegler:
Thanks @idkjs! Looks like I need to update some things with our Travis setup, the changes themselves look great!

PR #154, opened 2020-03-29
Support urql 1.4.x

JoviDeCroock:
This is my first attempt at Reason bindings, there could be some errors and trying to find out the questions that I listed.


src/UrqlClient.re
```diff
@@ -81,6 +79,9 @@ module ClientTypes = {
     partialOpUrl: string,
     [@bs.optional] [@bs.as "meta"]
     partialOpDebugMeta: operationDebugMeta,
+    // TODO: how to add this to useQuery...
```
JoviDeCroock: I was trying to find out how to only allow this in `useQuery`

parkerziegler: Gotcha. I don't think it's possible to restrict this property to _only_ be used by a `partialOperationContext` used by `useQuery` _and not_ `useMutation`. I think that's ok – it's something we can flag in docs.

To add this as an option that the `useQuery` hook can supply as part of the config rather than via a `partialOperationContext`, you can do the following:

1. Add `[@bs.optional] pollInterval: int` here: https://github.com/FormidableLabs/reason-urql/blob/master/src/hooks/UrqlUseQuery.re#L3-L12
2. Ensure you add an argument for it here, `pollInterval=?`: https://github.com/FormidableLabs/reason-urql/blob/master/src/hooks/UrqlUseQuery.re#L89
3. Make sure it gets forwarded along here: https://github.com/FormidableLabs/reason-urql/blob/master/src/hooks/UrqlUseQuery.re#L97 as `pollInterval?`

And then we'll want to follow the same steps for the `Query` component in `UrqlQuery.re`.

src/UrqlClient.re
```diff
@@ -260,6 +264,8 @@ type clientOptions('a, 'b) = {
   suspense: bool,
   [@bs.optional]
   fetch: 'b,
+  [@bs.optional]
+  requestPolicy: UrqlTypes.requestPolicy,
```
parkerziegler: Nice! We'll just need to make sure this gets exposed in a few more places. This `[@bs.deriving abstract]` binds the JS API, but we need to make sure the Reason API has a way for us to send this value along. Here are the steps we'll need to take:

1. Change this line to:

```suggestion
  [@bs.optional]
  requestPolicy: string,
```

In JS-land, this will be a `string`. The polymorphic variant we have for `requestPolicy` guarantees it'll never be a string other than those specified here: https://github.com/FormidableLabs/reason-urql/blob/master/src/UrqlTypes.re#L3-L9

2. Add an argument for `requestPolicy` here: https://github.com/FormidableLabs/reason-urql/blob/master/src/UrqlClient.re#L280, i.e. ~requestPolicy=`CacheFirst (sorry I can't code highlight here, the backtick for polymorphic variants is important). 
3. Convert from the polymorphic variant to the string representation here: https://github.com/FormidableLabs/reason-urql/blob/master/src/UrqlClient.re#L290 by adding: `~requestPolicy=requestPolicy->UrqlTypes.requestPolicyToJs`

This last piece ensures when we receive the polymorphic variant in Reason we turn it into a string on the JS-side.

src/UrqlClient.re
```diff
@@ -320,6 +326,8 @@ let partialOpCtxToPartialOpCtxJs = opts =>
   switch (opts) {
   | Some(o) =>
     let url = o->ClientTypes.partialOpUrlGet;
+    // TODO: where does this "partialOpPollIntervalGet" come from...
+    let pollInterval = o->ClientTypes.partialOpPollIntervalGet;
```
parkerziegler: Good question! It gets autogenerated by the use of `[@bs.deriving abstract]`: https://bucklescript.github.io/docs/en/object#accessors Since `@bs.deriving abstract` generates an "abstract" type, whose internal structure is hidden from the outside world, the only typesafe way to access properties is via getters and setters that are auto-generated by BuckleScript. It's not the most ergonomic but it guarantees you can't access fields that don't exist using dot notation, i.e. `o.someOtherPollInterval`.

JoviDeCroock:
Thank you so much @parkerziegler for adding the explanations to your review, this has been really awesome to learn more on how this all works. There seem to be a lot of moving parts, getting my head around it.

JoviDeCroock:
Somewhere the variant seems to be failing since

```
  describe("Client with default requestPolicy", () => {
    it("should accept a default requestPolicy", () => {
      let clent =
        Client.make(~url="https://localhost:3000", ~requestPolicy=`cacheFirst, ());
    });
  });
```

Results in

```
  This has type:
    [> `cacheFirst ]
  But somewhere wanted:
    UrqlTypes.requestPolicy
```

parkerziegler:
No problem @JoviDeCroock, thanks for the PR! You just need to capitalize that polymorphic variant, from `cacheFirst to `CacheFirst.

__tests__/__snapshots__/UrqlClient_test.bs.js.snap
```diff
@@ -1,5 +1,26 @@
...
+  "dispatchOperation": [Function],
+  "exchange": [Function],
+  "executeMutation": [Function],
+  "executeQuery": [Function],
+  "executeSubscription": [Function],
+  "fetch": undefined,
+  "fetchOptions": undefined,
+  "operations$": [Function],
+  "reexecuteOperation": [Function],
+  "requestPolicy": "cache-and-network",
```
parkerziegler: 🎉 Just wanted to ensure that a different policy other than `cache-first` got picked up since, IIRC, that's already the default.

parkerziegler:
This looks ready to go @JoviDeCroock, thanks for hopping in and getting your hands dirty with Reason and BuckleScript! Any changes to support 1.4 other than the addition of `pollInterval` and a default `requestPolicy` on the Client? If not, I'll add some docs for this tomorrow and cut a release!

parkerziegler:
@all-contributors please add @JoviDeCroock for code.

JoviDeCroock:
Judging from the changelog this should cover 1.4.x, I’ll probably try to incrementally send PR’s until we’re up to date, that way people using the lib don’t have to force the upgrade and we have a nice git trail.

parkerziegler:
Sounds great, I'll be able to help with reviews and code as well, so hopefully together we can patch this up in no time. I'll put a PR up for docs this afternoon (my time) and get the changelog ready to go.

PR #157, opened 2020-04-06
Add support for promisified query and mutation methods.

parkerziegler:
This PR begins adding support for `urql` v1.5, namely the promisified `query` and `mutation` methods on the Client 🎉 This is fortunately quite easy in Reason, as the TS implementations of this just call `Wonka.take` and `Wonka.toPromise` under the hood.

As part of housekeeping, I also upgraded our BuckleScript version to 7 (finally), changed our default project compilation to ES6, and added in the proper transform to support testing, similar to what we have in [`renature`](https://github.com/FormidableLabs/renature/blob/master/scripts/jest-transform-esm.js) and what @kitten showed in his commit on the `v2` branch.

Next task will be adding support for the `useClient` hook, and then I think we'll have what we need to cut a release 🎉

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   "dependencies": {
     "graphql": "^14.1.1",
     "urql": "1.5.1",
-    "wonka": "^4.0.9"
+    "wonka": "^3.2.2"
   },
   "devDependencies": {
     "@babel/plugin-transform-modules-commonjs": "^7.9.0",
--- a/package.json
+++ b/package.json
@@ -27,26 +27,36 @@
   "license": "MIT",
   "dependencies": {
     "graphql": "^14.1.1",
-    "urql": "1.4.1",
-    "wonka": "^3.1.0"
+    "urql": "1.5.1",
+    "wonka": "^4.0.9"
   },
   "devDependencies": {
-    "@glennsl/bs-jest": "^0.4.8",
+    "@babel/plugin-transform-modules-commonjs": "^7.9.0",
+    "@glennsl/bs-jest": "^0.5.1",
     "all-contributors-cli": "^5.4.0",
-    "bs-fetch": "^0.5.0",
-    "bs-platform": "6.2.1",
+    "babel-jest": "^25.2.6",
+    "bs-fetch": "^0.5.2",
+    "bs-platform": "7.2.2",
+    "gentype": "^3.18.0",
     "husky": "^2.4.1",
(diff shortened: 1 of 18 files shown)


parkerziegler:
Yeah, intersection types in Reason are quite difficult. In the case of most `wonka`-related things we have the benefit of it already being implemented in Reason, so we can definitely use that to our advantage.

package.json
```diff
@@ -27,26 +27,36 @@
...
+    "bs-platform": "7.2.2",
+    "gentype": "^3.18.0",
     "husky": "^2.4.1",
     "lint-staged": "^8.2.1",
     "npm-run-all": "^4.1.5",
     "reason-react": "0.7.0"
   },
   "peerDependencies": {
-    "bs-fetch": "^0.5.0",
+    "bs-fetch": "^0.5.2",
```
kitten, line 47: On a side note, why is this a peer dependency? I was trying to figure that out but I’m not sure whether having it as just a dependency would cause issues?

src/UrqlClient.re
```diff
@@ -486,3 +486,51 @@ let createRequestOperation = (~client, ~operationType, ~request, ~opts=?, ()) =>
...
+let query =
+    (
+      ~client: t,
+      ~request: UrqlTypes.request('response),
+      ~opts: option(ClientTypes.partialOperationContext)=?,
+      (),
+    ) => {
+  executeQuery(~client, ~request, ~opts?, ())
+  |> Wonka.take(1)
+  |> Wonka.toPromise;
```
kitten, line 511: Does it make more sense maybe not to convert these to promises? It seems to me like you wouldn’t gain much in Reason from using promises 😅

package.json
parkerziegler: I don't think it would. It's a dep so we can get proper type support for `Fetch.requestInit`. But yeah, I think we could just make it a dependency and save everyone the trouble.

src/UrqlClient.re
parkerziegler: Agreed that I don't think we necessarily gain much, but I think it's still good to cover in the lib in case people want to interact with the `Js.Promise.t` type rather than `Wonka.sourceT`. I could imagine folks coming from JS and learning about `Js.Promise.t` and wanting a way to get a `Js.Promise.t` from these methods. These bindings are such low cost anyway that I think it's still worthwhile (tho I prefer `Wonka.subscribe` personally as well).

PR #160, opened 2020-04-11
Add stale and suspense

JoviDeCroock:
After this addition we should be able to upgrade straight up to 1.9.x

For later @urql/core versions we still need:
- https://github.com/FormidableLabs/urql/blob/master/packages/core/CHANGELOG.md#190
- https://github.com/FormidableLabs/urql/blob/master/packages/core/CHANGELOG.md#1100


parkerziegler:
Closing in favor of #184.

PR #161, opened 2020-04-13
Add bindings for client.readQuery

tatchi:
This add bindings for `client.readQuery`


src/UrqlClient.re
```diff
@@ -511,6 +511,32 @@ let query =
...
+[@bs.send]
+external readQueryJs:
+  (
+    ~client: t,
+    ~query: string,
+    ~variables: Js.Json.t=?,
+    ~context: ClientTypes.partialOperationContextJs=?,
+    unit
+  ) =>
+  Js.Null.t('response) =
```
tatchi, line 523: I was not sure between `Js.Null.t` and `Js.Nullable.t`. I chose the former since it should reflect better the return type from the js side: https://github.com/FormidableLabs/urql/blob/master/packages/core/src/client.ts#L266

parkerziegler: ✅ I think that makes sense. AFAIK, `Js.Nullable.t` is generally preferred to handle both `null` and `undefined` but since, as you point out, the JS side explicitly uses `null` I think this looks great.

src/UrqlClient.re
```diff
@@ -511,6 +511,32 @@ let query =
   |> Wonka.toPromise;
 };
 
+[@bs.send]
+external readQueryJs:
```
parkerziegler, line 515: So in this case w/ `readQueryJs` you seem to be writing a binding for `readQuery` but never actually using it; instead, `readQuery` below just re-implements the same method in Reason. I actually think the latter is nicer – it takes advantage of the niceties of Reason to do what `pipe` does in `urql`.

TL;DR I think you can get rid of L514 – 524 altogether 😄

parkerziegler:
👋 Thanks @tatchi for this, it looks really nice. The only concern I have is that we're binding to a much earlier version of `urql` currently (v1.5.1). So your `[@bs.send]` binding wouldn't actually work, b/c prior to v1.9.0 the `readQuery` method didn't exist on the client 😬 I'm still happy to keep the PR around, however, and we can merge it once we get the bindings up to speed there.

Removing the `[@bs.send]` binding as I commented would fix the binding issue, but also mean we introduce a feature earlier than expected in the bindings, which I'm very wary of doing. If you're up to making the discussed changes I'd be happy to merge this once v1.9.0 is slated for bindings.

src/UrqlClient.re
tatchi: Yes that makes sense. Actually I've been influenced by the `queryJs` and `mutationJs` ones. If I'm not mistaken, those are not used either. What's the reason to keep them then? Is it because one version returns a `Wonka.sourceT` while the other returns a `promise`?

tatchi:
> 👋 Thanks @tatchi for this, it looks really nice. The only concern I have is that we're binding to a much earlier version of `urql` currently (v1.5.1). So your `[@bs.send]` binding wouldn't actually work, b/c prior to v1.9.0 the `readQuery` method didn't exist on the client 😬 I'm still happy to keep the PR around, however, and we can merge it once we get the bindings up to speed there.
> 
> Removing the `[@bs.send]` binding as I commented would fix the binding issue, but also mean we introduce a feature earlier than expected in the bindings, which I'm very wary of doing. If you're up to making the discussed changes I'd be happy to merge this once v1.9.0 is slated for bindings.

Thanks for the review! 

Yes, that totally makes sense to wait for 1.9.0 before merging this. If I understand it correctly it's probably coming soon thanks to #160

src/UrqlClient.re
parkerziegler: Ah that's actually a good shout, they can be removed as well! Good eye 👀

tatchi:
Hey, thanks for the update. I'm indeed interested to finish this one 👍 I'll give it a shot in the coming days and will you know once it's done :)

tatchi:
@parkerziegler Should be ok :)

src/Client.re
```diff
@@ -397,3 +397,30 @@ let mutation =
...
+
+let readQuery =
+    (
+      ~client,
+      ~request,
+      ~fetchOptions=?,
+      ~requestPolicy=?,
+      ~url=?,
+      ~pollInterval=?,
+      ~meta=?,
```
parkerziegler, line 409: One thing to note – we recently added support for the full spectrum of options that can be passed in a `Partial<OperationContext>`. These new additions include `fetch` (custom `fetch` implementation to use), `additionalTypenames` (let the operation know that it depends on specific `__typenames`, used mostly by the standard document cache), `suspense` (a boolean flag to enable suspense), and `preferGetMethod` (instruct GraphQL operations to be executed with `GET` rather than `POST` as the http method).

You can look at `executeQuery` and match that signature (since it's the same) and we should be good to go! I'm also realizing I need to update the bindings for `query` and `mutation` while I'm here as well.

parkerziegler:
Looking great @tatchi, one last set of changes then we are good to merge 🚀

src/Client.re
tatchi: Thanks for the detailed explanation 😊  I indeed took inspiration from the `query` one so I missed those extra options

tatchi:
Thanks for the review @parkerziegler. I opened #203 to update the `query` and `mutation`

parkerziegler:
@all-contributors please add @tatchi for code.

PR #165, opened 2020-04-19
Remove useMemo from useSubscription

jeddeloh:
This is more of an issue than a PR. `useSubscription` is currently wrapping the JS hook with `useMemo` which errors with "can't call a hook from within a hook". It wasn't immediately apparent to me why `useMemo` was being used in quite this way. I would have expected the JS component to be doing proper checks on the args provided to it and `useSubscriptionArgs` looks to me like it would create a new object on every render anyway defeating `useMemo`? Anyway, I didn't dig too deep and just removed `useMemo` here. I'm sure there was a reason I'm not seeing—maybe it's worth a comment once it gets figured out.

Thanks!

### DIFF
--- a/src/hooks/UrqlUseSubscription.re
+++ b/src/hooks/UrqlUseSubscription.re
@@ -81,23 +81,18 @@ let useSubscription =
       (),
     );
 
-  React.useMemo3(
-    () => {
-      let response: UrqlTypes.hookResponse(ret, 'extensions) =
-        switch (handler) {
-        | Handler(handlerFn) =>
-          useSubscriptionJs(
-            args,
-            Some((acc, data) => handlerFn(acc, parse(data))),
-          )[0]
-          |> useSubscriptionResponseToRecord(x => x)
-        | NoHandler =>
-          useSubscriptionJs(args, None)[0]
-          |> useSubscriptionResponseToRecord(parse)
-        };
+  let response: UrqlTypes.hookResponse(ret, 'extensions) =
+    switch (handler) {
+    | Handler(handlerFn) =>
+      useSubscriptionJs(
+        args,
+        Some((acc, data) => handlerFn(acc, parse(data))),
+      )[0]
+      |> useSubscriptionResponseToRecord(x => x)
+    | NoHandler =>
+      useSubscriptionJs(args, None)[0]
+      |> useSubscriptionResponseToRecord(parse)
+    };
 
-      response;
-    },
-    (handler, args, parse),
-  );
+  response;
 };
--- a/src/hooks/UrqlUseSubscription.re
+++ b/src/hooks/UrqlUseSubscription.re
@@ -29,7 +29,7 @@ external useSubscriptionJs:
  * A function for converting the response to useQuery from the JavaScript
  * representation to a typed Reason record.
  */
-let jsSubscriptionResponseToRecord =
+let urqlResponseToReason =
     (result): UrqlTypes.hookResponse('response, 'extensions) => {
   let data = result->UrqlTypes.jsDataGet->Js.Nullable.toOption;
   let error =
@@ -90,8 +90,5 @@ let useSubscription =
 
   let (jsResponse, _) = useSubscriptionJs(args, Some(handler'));
 
-  UrqlGuaranteedMemo.useGuaranteedMemo1(
-    jsSubscriptionResponseToRecord,
-    jsResponse,
-  );
+  UrqlGuaranteedMemo.useGuaranteedMemo1(urqlResponseToReason, jsResponse);
 };
--- a/src/hooks/UrqlUseSubscription.re
+++ b/src/hooks/UrqlUseSubscription.re
@@ -5,7 +5,7 @@
  */
 type handler('acc, 'resp, 'ret) =
   | Handler((option('acc), 'resp) => 'acc): handler('acc, 'resp, 'acc)
-  | NoHandler: handler(_, 'resp, 'resp);
+  | NoHandler: handler('resp, 'resp, 'resp);
 
 /* Arguments passed to useSubscription on the JavaScript side. */
 [@bs.deriving abstract]
@@ -60,8 +60,7 @@ let jsSubscriptionResponseToRecord =
  * to the GraphQL subscription, and a parse function for decoding the JSON response.
  *
  * handler – an optional function to accumulate subscription responses.
- */;
-
+ */
 let useSubscription =
     (
       type acc,
@@ -83,14 +82,13 @@ let useSubscription =
       (),
     );
 
-  let handler' =
+  let handler' = (acc, jsData) =>
     switch (handler) {
-    | Handler(handlerFn) =>
-      Some((acc, data) => handlerFn(acc, parse(data)))
-    | NoHandler => Some((_acc, data) => Obj.magic(parse(data)))
+    | Handler(handlerFn) => handlerFn(acc, parse(jsData))
+    | NoHandler => parse(jsData)
     };
 
-  let (jsResponse, _) = useSubscriptionJs(args, handler');
+  let (jsResponse, _) = useSubscriptionJs(args, Some(handler'));
 
   UrqlGuaranteedMemo.useGuaranteedMemo1(
     jsSubscriptionResponseToRecord,
--- a/src/hooks/UrqlUseSubscription.re
+++ b/src/hooks/UrqlUseSubscription.re
@@ -17,20 +17,21 @@ type useSubscriptionArgs = {
   context: UrqlClient.ClientTypes.partialOperationContext,
 };
 
+type executeSubscriptionJs;
+
 [@bs.module "urql"]
 external useSubscriptionJs:
   (useSubscriptionArgs, option((option('acc), Js.Json.t) => 'acc)) =>
-  array(UrqlTypes.jsResponse('ret, 'extensions)) =
+  (UrqlTypes.jsResponse('ret, 'extensions), executeSubscriptionJs) =
   "useSubscription";
 
 /**
  * A function for converting the response to useQuery from the JavaScript
  * representation to a typed Reason record.
  */
-let useSubscriptionResponseToRecord =
-    (parse, result): UrqlTypes.hookResponse('response, 'extensions) => {
-  let data =
-    result->UrqlTypes.jsDataGet->Js.Nullable.toOption->Belt.Option.map(parse);
+let jsSubscriptionResponseToRecord =
+    (result): UrqlTypes.hookResponse('response, 'extensions) => {
+  let data = result->UrqlTypes.jsDataGet->Js.Nullable.toOption;
   let error =
     result
     ->UrqlTypes.jsErrorGet
@@ -60,6 +61,7 @@ let useSubscriptionResponseToRecord =
  *
  * handler – an optional function to accumulate subscription responses.
  */;
+
 let useSubscription =
     (
       type acc,
@@ -81,18 +83,17 @@ let useSubscription =
       (),
     );
 
-  let response: UrqlTypes.hookResponse(ret, 'extensions) =
+  let handler' =
     switch (handler) {
     | Handler(handlerFn) =>
-      useSubscriptionJs(
-        args,
-        Some((acc, data) => handlerFn(acc, parse(data))),
-      )[0]
-      |> useSubscriptionResponseToRecord(x => x)
-    | NoHandler =>
-      useSubscriptionJs(args, None)[0]
-      |> useSubscriptionResponseToRecord(parse)
+      Some((acc, data) => handlerFn(acc, parse(data)))
+    | NoHandler => Some((_acc, data) => Obj.magic(parse(data)))
     };
 
-  response;
+  let (jsResponse, _) = useSubscriptionJs(args, handler');
+
+  UrqlGuaranteedMemo.useGuaranteedMemo1(
+    jsSubscriptionResponseToRecord,
+    jsResponse,
+  );
 };
--- a/src/utils/UrqlGuaranteedMemo.re
+++ b/src/utils/UrqlGuaranteedMemo.re
@@ -1,4 +1,4 @@
-// React.useMemo has to semantic guarantee so we define our own
+// React.useMemo has no semantic guarantee so we define our own
 let useGuaranteedMemo1 = (computeValue, arg) => {
   let (memoizedState, setMemoizedState) =
     React.useState(() => computeValue(arg));
--- a/src/utils/UrqlGuaranteedMemo.re
+++ b/src/utils/UrqlGuaranteedMemo.re
@@ -1,0 +1,15 @@
+// React.useMemo has to semantic guarantee so we define our own
+let useGuaranteedMemo1 = (computeValue, arg) => {
+  let (memoizedState, setMemoizedState) =
+    React.useState(() => computeValue(arg));
+
+  React.useEffect1(
+    () => {
+      setMemoizedState(_ => computeValue(arg));
+      None;
+    },
+    [|arg|],
+  );
+
+  memoizedState;
+};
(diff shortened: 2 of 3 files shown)


gugahoa:
That error is a bit strange, afaik [you can call hooks from custom hooks](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-from-react-functions) which this function is.

The reason for the `useMemo`  can be found here https://github.com/FormidableLabs/reason-urql/pull/83:

> Problem 2
>
>The return value from reason-urql hooks were not safe to use in the dependency list of useEffect, useMemo, useCallback and similar, even though the response JS-land seems safe to use. That's caused by us converting the Js object to an Reason type, and at runtime that always creates a new object

Previously it was like what's proposed here (you can see it here [`f4b9dd6`](https://github.com/FormidableLabs/reason-urql/pull/83/commits/f4b9dd693fc07a83313026f541ba1ef364a9908b#diff-d6d154cd53e80127bed7c60be46a2d05))

There were other places that didn't have `useMemo` but that PR introduced too, for the same reasons.

That said, it's still strange that you got an error about not being able to use the hook from within a hook. Could you send a reproduction repo with that error?

jeddeloh:
Awesome! Thanks for the info and quick response.

#### Hooks within Hooks
Maybe we can avoid a repro if we can align on how hooks work?   It’s my understanding that a function becomes a custom hook as soon as it calls another hook. Before that it was just a function. However, this code is not calling this hook directly from a _custom hook_, it's calling it from _within a real hook_.

I think this is a violation of **“Only Call Hooks at the Top Level“**

Excerpt from the Rules of Hooks you posted:
> Don’t call Hooks inside loops, conditions, or nested functions. Instead, always use Hooks at the top level of your React function. By following this rule, you ensure that Hooks are called in the same order each time a component renders. That’s what allows React to correctly preserve the state of Hooks between multiple useState and useEffect calls. (If you’re curious, we’ll explain this in depth below.)

So `useMemo` is kinda like a conditional. Some renders it will call the hook, some renders it won’t. Even if you got it to pass here, I would expect it to eventually error with the “rendered with a different number of hooks” error. When that’s going on I don’t see how React can keep track of which hook is which.

#### UseMemo
Regarding the intent for the return value to be “safe”, is the intent that when given to useEffect it should run once and only once when the response changes? If so, I think something else should be used entirely since **“You may rely on useMemo as a performance optimization, not as a semantic guarantee.”**: https://reactjs.org/docs/hooks-reference.html#usememo

Is there a specific use-case where you would want to referentially check the whole response? I think I would always be checking specific values to determine if an effect should be run, but maybe this is worth the effort.

#### How to move forward
I actually can’t even figure out how this code is working (I mean that sincerely—not trying to be snotty). When new data comes in on your subscription, `useSubscriptionJs` should update a `useState` hook internally somewhere which should trigger a new render. However we’re memoizing on… `handler, args, parse`? No response? Theoretically, (if we could rely on useMemo as a semantic guarantee) `useMemo` would run only the first time and we would never return a new response unless `handler`, `args`, or `parse` changed which doesn't seem like the intent. Luckily, it looks to me like we create a new args on every render so `useMemo` is not actually doing anything and should be running on every render… :shrug:

I guess if the problem is in converting the response to record I would expect some sort of check on response changed and only create new value at that time like:
```
  React.useMemo1(
    () => {
      response->subscriptionResponseToRecord;
    },
    [|response|],
  );
```
But use something that is guaranteed to run once instead of `useMemo` :) Maybe I’m missing something in my analysis. 

Thanks!

gugahoa:
> However, this code is not calling this hook directly from a __custom hook__, it's calling it from __within a real hook__.

That makes sense, I hadn't thought about that
> Regarding the intent for the return value to be “safe”, is the intent 
that when given to useEffect it should run once and only once when the 
response changes?

Yes
> Is there a specific use-case where you would want to referentially check
 the whole response? I think I would always be checking specific values 
to determine if an effect should be run, but maybe this is worth the 
effort.

This would also happen when you check specific values, for example, when the specific value is an Array, afaik it'll use referential check to know if the Array changed or not

> I actually can’t even figure out how this code is working
From your analysis, I also can't seem to figure it out. Seems like it's working by chance.

Your solution seems to be what I should have gone for. But you said to use something other than `useMemo`, what would you use?

jeddeloh:
>Your solution seems to be what I should have gone for. But you said to use something other than useMemo, what would you use?

I think you might just have to build it up from basic building blocks:
```
 let useSemanticallyGuaranteedMemo1 = (computeValue, arg) => {
  let (memoizedState, setMemoizedState) =
    React.useState(() => computeValue(arg));

  React.useEffect1(
    () => {
      setMemoizedState(_ => computeValue(arg));
      None;
    },
    [|arg|],
  );

  memoizedState;
};
```

  That aside, I’m trying to understand the use-case for memoizing in the first place. Here’s the shape of the hook response:
```
 type hookResponse('ret, 'extensions) = {
  fetching: bool,
  data: option('ret),
  error: option(UrqlCombinedError.t),
  response: response('ret),
  extensions: option('extensions),
}; 
```

We are trying to memoize the _data_  or data in the _response_, not the whole hookResponse, right? I can't think of a use case to be checking the actual record with fetching, error, etc.

  If we’re looking at data (at one spot or another) there's two situations to consider:

- a render is going on that was not driven by an actual change (react apparently just does this sometimes)
- an actual change has occured

In the no-change case the useSubscription hook in javascript should be returning a data value that is referentially equal to the one on the last render. If not, then it seems like a bug in Urql.

In the case where a change _has_ happened, we (the library authors) have no idea what the shape of the data is, all we can do is a referential check on the data object itself and triggering a change here makes total sense to me.

> This would also happen when you check specific values, for example, when the specific value is an Array, afaik it'll use referential check to know if the Array changed or not

Let’s say your subscription is returning data like: 
```
 {   
  .
  chatRoomUsers: array(string),
  comments: array(string),
}
```

And you’re in the situation where `chatRoomUsers` has changed, but comments have not, and you don’t want to rerender comments in this case. Seems like here there are two considerations again:

- Your subscription is returning a whole list of comments with every update
- You're using a handler on `useSubscription` to add comments one at a time to the comments value

In the case your subscription is returning a whole list of comments with every update, that seems expected and unavoidable. You (as the user) could implement your own comparator function to see if the _contents_ of two arrays are equal if you really needed that. Maybe length check or something would be sufficient.

In the other case of using a handler on `useSubscription` to add comments one at a time to the comments value, I would assume it depends on how you manipulate the previous value, but I don’t know. If that’s the issue, we could dig in further.

Aside from possibly that one last point, I’m kinda of the opinion that the useMemo shouldn’t be needed. Thoughts?

gugahoa:
> We are trying to memoize the data or data in the response, not the whole hookResponse, right? I can't think of a use case to be checking the actual record with fetching, error, etc.

I believe we should memoize the whole response, and re-calculate based on a referential check to the return values from the JS hook. Then the hook behavior will be quite similar to the original useSubscription behavior.

> In the no-change case the useSubscription hook in javascript should be returning a data value that is referentially equal to the one on the last render. If not, then it seems like a bug in Urql.

When that code was merged, useSubscription in urql behaved correctly, and internally maintains its own state, I believe that hasn't changed.

> Aside from possibly that one last point, I’m kinda of the opinion that the useMemo shouldn’t be needed. Thoughts?  

I'm with you here too. Maybe it's also a good chance to check the other uses of useMemo in the same PR, and create issues if they're not being properly used too.

jeddeloh:
I pushed up a pass at memoizing the response based on what is returned from JS.

src/hooks/UrqlUseSubscription.re
```diff
@@ -81,23 +83,17 @@ let useSubscription =
...
-          )[0]
-          |> useSubscriptionResponseToRecord(x => x)
-        | NoHandler =>
-          useSubscriptionJs(args, None)[0]
-          |> useSubscriptionResponseToRecord(parse)
-        };
+  let handler' =
+    switch (handler) {
+    | Handler(handlerFn) =>
+      Some((acc, data) => handlerFn(acc, parse(data)))
```
jeddeloh, line 89: I opted to always have the data coming out of `useSubscriptionJs` be consistent by passing a handler that will parse by default. However, I had to `Obj.magic` here because the types seemed to enforce that `acc` and `resp` be different. I started tracking it down but things didn't makes sense. `type handler` takes `'ret` but is not used?

src/hooks/UrqlUseSubscription.re
```diff
@@ -17,20 +17,21 @@ type useSubscriptionArgs = {
   context: UrqlClient.ClientTypes.partialOperationContext,
 };
 
+type executeSubscriptionJs;
+
 [@bs.module "urql"]
 external useSubscriptionJs:
   (useSubscriptionArgs, option((option('acc), Js.Json.t) => 'acc)) =>
-  array(UrqlTypes.jsResponse('ret, 'extensions)) =
+  (UrqlTypes.jsResponse('ret, 'extensions), executeSubscriptionJs) =
```
jeddeloh, line 25: The Urql documentation seemed to indicate this was a tuple, and I think this is okay as arrays are the underlying representation of tuples.

src/hooks/UrqlUseSubscription.re
gugahoa: Could you elaborate a bit on why you opted that way? Is it to not pass `parse` as argument to `jsSubscriptionResponseToRecord`? Which benefits do you see?

The reason why the compiler expects resp and acc to be different is because of how `handler('acc, 'resp, 'ret)` has been declared (It's a [GADT](https://sketch.sh/s/yH0MJiujNSiofDWOU85loX/))

So what's happening here is that in the first branch `'acc` and `'ret` are the same, guaranteed by `handler('acc, 'resp, 'ret)` type, but for the second branch no relation is guaranteed between `'acc` and `'ret`

The type `'ret` in `handler('acc, 'resp, 'ret)` is there only to help the compiler know what's the return value for the `useSubscription` when passing a `Handler` and when passing a `NoHandler`

GADTs can be a little confusing, if you wish to, I'll hang out on Spectrum where we can chat in real time

src/hooks/UrqlUseSubscription.re
gugahoa: I believe it's ok too, you could even only have `(UrqlTypes.jsResponse('ret, 'extensions), 'executeSubscriptionJs)` instead of declaring a concrete type for `executeSubscriptionJs` since we aren't using it anyway

src/hooks/UrqlUseSubscription.re
jeddeloh: > Could you elaborate a bit on why you opted that way? Is it to not pass parse as argument to jsSubscriptionResponseToRecord? Which benefits do you see?

I guess my reasoning is something like this:

1. In terms of reasoning about hooks, calling them inside of a switch is much like a conditional. We're calling the same one so it works, but as a matter of principle I'd like to try and compute the arguments once and only call the hook once so I don't fall into a trap.
2. The fact that previously the result of one call to `useSubscription` had to be parsed and one with a handler did not seemed odd to me. I'd like the return values to be used in the same way. I'd also like to be parsing in the same way, in the same place. If you think of `handler` more like a transform rather than a function to mash new data on to the previous value, this seemed reasonable. The transform for `NoHandler` is identity. Now we can parse the data going into the transform consistently for both cases.
3. Since we had to memoize the conversion to a record, having a single definition of the function was just simpler. (I also removed `use` from the function name since it is not actually a hook)

> So what's happening here is that in the first branch 'acc and 'ret are the same, guaranteed by handler('acc, 'resp, 'ret) type, but for the second branch no relation is guaranteed between 'acc and 'ret

Ah! I see what you're doing now. Thanks for that! I switched `NoHandler` to be `handler('resp, 'resp, 'resp)` (don't ask to infer the `'acc` type) and it works now. Pushing up a change. A lot of changes that maybe weren't necessary here, but made it easier for me to reason about. I'm totally open to changing it if you prefer a solution closer to how it was previously.

src/hooks/UrqlUseSubscription.re
jeddeloh: I think if we were to use a type variable here, and someone were to actually use it, the compiler would just infer it to be whatever it was at the call site and make the program unsafe. While unlikely, my reasoning was that it should be clear that it's not typed yet. Someone could always cast to whatever they need with `Obj.magic` or `%identity` if they needed it in a hurry.

gugahoa: Sounds perfectly fine to me then!

src/hooks/UrqlUseSubscription.re
gugahoa: Your changes are just fine, and I also agree with you that it made it easier to reason about! :)

src/hooks/UrqlUseSubscription.re
parkerziegler, line 25: One thing to note here – this type can actually be expressed non-abstractly. It's the `executeSubscription` function, which is unfortunately not well documented in the `urql` docs. You have to look at the source for the version we're binding to (currently 1.5.1) to know: https://github.com/FormidableLabs/urql/blob/v1.5.1/src/hooks/useSubscription.ts#L55-L80.

The use cases for `executeSubscription` in the wild aren't many; I chatted with Phil Pluckthun about this today (also on `urql` core team) and it's mostly for feature parity with the other hooks. All this being said, I'll tackle this in a follow up PR; more just a good thing to know about, and thanks for finding this and bringing it up.

src/hooks/UrqlUseSubscription.re
```diff
@@ -17,20 +17,21 @@ type useSubscriptionArgs = {
...
 
 /**
  * A function for converting the response to useQuery from the JavaScript
  * representation to a typed Reason record.
  */
-let useSubscriptionResponseToRecord =
-    (parse, result): UrqlTypes.hookResponse('response, 'extensions) => {
-  let data =
-    result->UrqlTypes.jsDataGet->Js.Nullable.toOption->Belt.Option.map(parse);
+let jsSubscriptionResponseToRecord =
```
parkerziegler, line 32: We do have a consistent name for this function in other modules:

```suggestion
let urqlResponseToReason =
```

Mind updating it just to keep the nomenclature consistent and scannable across these hooks?

src/utils/UrqlGuaranteedMemo.re
```diff
@@ -0,0 +1,15 @@
+// React.useMemo has to semantic guarantee so we define our own
```
parkerziegler, line 1:
```suggestion
// React.useMemo has no semantic guarantee so we define our own
```

src/utils/UrqlGuaranteedMemo.re
```diff
@@ -0,0 +1,15 @@
...
+let useGuaranteedMemo1 = (computeValue, arg) => {
+  let (memoizedState, setMemoizedState) =
```
parkerziegler, line 3: Nice, this looks great 🚀 I'm down to include this in our `hooks` folder since I envision we may want to use this in other places. I am trying to think through the implications of this value being tracked in state. Upstream in `urql`'s `useSubscription`, subscription data is already kept in `state`, meaning we get a re-render every time a new message comes in, which would run this hook. I'm wondering if the `state` update here will then trigger _another_ re-render. I'm assuming React would catch them, but wondering if we may want to use a `ref` here instead. Let's keep it as is for now, and I'll do some experiments to see if we're getting additional unnecessary renders.

src/hooks/UrqlUseSubscription.re
jeddeloh: 👍

src/utils/UrqlGuaranteedMemo.re
jeddeloh: Please do! I did not consider all that carefully if this is the best approach or if there are patterns already established. I also initially considered using a ref, but didn't immediately see a way to make it work since useEffect is async and you need a way to trigger a re-render after it runs.

jeddeloh:
I made those small updates so we should be good to go. Thanks!

src/utils/UrqlGuaranteedMemo.re
jeddeloh: Actually, now that I am thinking about it, perhaps `useCallback` is the better solution? I don't see any restrictions about the semantics of `useCallback` in the React documentation.
```
  let memoizedUrqlResponseToReason = React.useCallback1(
    () => urqlResponseToReason(jsResponse),
    [|jsResponse|]
  );

  memoizedUrqlResponseToReason()
```

parkerziegler:
Thanks again!

parkerziegler:
@all-contributors please add @jeddeloh for code.

Issue #144, opened 2020-02-06
[Proposal] Init value in response type


gaku-sei:
The logic here https://github.com/FormidableLabs/reason-urql/blob/8bb9cf6de62cc3b870bb7b7e59c79f136f04c0ca/src/components/UrqlMutation.re#L59 looks a bit off for dynamic mutations. 

By definition, a dynamic mutation will occur after the hook is called, so `fetching` will always be `false`, and `data` and `error` will always be `None` as long as the mutation hasn't been called.

I propose to introduce an `Init` (or something similar) to handle this case in the response type. This will very likely require the use of a state to know whether or not the mutation has been triggered already or not.

We could even consider having a `Refetching` constructor if needed.

The Relude library handles this case as well: https://github.com/reazen/relude/blob/master/src/Relude_AsyncData.re#L11

parkerziegler:
🤔 Hmmm, I'm not sure I fully understand the use case here. How does the `Init` constructor differ from `NotFound`? `NotFound` essentially acts as a catch all case, and can handle both:

1) The case you mention, where you haven't yet executed a mutation.
2) The case where you execute a mutation and, for some reason, you receive neither data or errors.

I'm not sure what `Refetching` would add either. It's not supported upstream in `urql` to differentiate loading states between "this is the first fetch" and "this is a subsequent fetch", so I'm not inclined to support that in the bindings. It's also fairly easy, as you mention, for a user to track whether or not this particular execution is the first execution simply using their own `useState` or `useRef` hook that updates when they call `executeMutation`.

If you can provide a more fleshed out use case for `Init` that shows how it substantively differs from `NotFound`, I'd consider adding it in. For now, the core focus of the library is on binding `urql` as closely as possible, and `NotFound` represents a "catch all" case we provide as a combination of initial state and a truly "no data or errors found" state. Hope that makes sense!

gaku-sei:
Thank you very much for the quick reply.

There are two different use cases here, one for `Init` and one for `Refetching`.

### Init

The way I understand `NotFound` is the one you're describing: "The Urql client hasn't found any data for this request".

But there is a difference between "The client hasn't found any data", and "The client hasn't found any data _yet_". Conceptually, those are two different things, and one might need to display two different ui: "Please perform a request" and "The request has been performed, but nothing was found". It's actually a situation I had to face pretty often, and it doesn't really sound like an edge case.

[...]

PR #168, opened 2020-05-16
🎉 v2

parkerziegler:
Alright, this is an exciting one – `v2`! As `urql` has grown in popularity in JS-land, getting these bindings upgraded feels more important than ever. However, it was time for a major overhaul to get rid of all our `[@bs.deriving abstract]` logic in favor of [direct record -> object compilation](https://reasonml.org/docs/reason-compiler/latest/object#records-as-objects). I also wanted to make the repo more modular and add proper `.rei` files for each module. Finally, I wanted to upgrade and migrate all our examples to latest `reason-react` and `graphql_ppx` and ensure they still compile and function with upgraded deps.

Overall, I think we're pretty much there! If you have a moment, I'd love any and all review you'd be willing to give. Most of the public-facing API changes are quite small; lots of internal work instead here. I'd love any thoughts on #144 as well; I didn't add it here but I think it may be useful and would want to add it in if others agree.

I'll be upgrading docs and CHANGELOG on this branch as we review. Once this gets merged we can resume the `urql` version upgrades and begin strategizing around bindings for larger pieces (like `graphcache`).


examples/4-exchanges/src/index.re
```diff
@@ -1,13 +1,10 @@
...
 
 /* This is the native debugExchange that ships with `urql`, re-implemented in Reason.
      Typically, you'd just add Exchanges.debugExchange to the Client's exchange array.
    */
-let debugExchange = exchangeInput => {
-  let forward = Exchanges.forwardGet(exchangeInput);
-
-  ops =>
+let debugExchange = (Exchanges.{forward}) =>
+  (. ops) =>
```
parkerziegler, line 7: The signature for exchanges had to change a bit. It was a happy accident that this was working before, made possible by the use of `forwardGet`, which forced compilation into a higher order function. But we really need to use uncurrying here to ensure the exchange actually returns a function that accepts operations and passes them through.

src/client/UrqlClientTypes.re
```diff
@@ -0,0 +1,143 @@
...
+  [@bs.optional]
+  url: string,
+  [@bs.optional]
+  debugMeta: operationDebugMeta,
+  [@bs.optional]
+  pollInterval: int,
+};
+
+module PartialOperationContextJs = {
+  [@bs.deriving {abstract: light}]
```
parkerziegler, line 64: Sadly had to keep around this `[@bs.deriving abstract]`. This is because we can't have partial objects with records; all fields on the record need to be explicitly specified. This meant that we'd need to create records for `partialOperationContext` with `None` for all fields that the user didn't want to modify, which then explicitly sets these to `undefined`. When `urql` receives these it'll overwrite the defaults with `undefined`, which blows up operations 😭 

`abstract: light` just helps us get tighter names for the getters and setters generated by BuckleScript.

parkerziegler:
Merging this now. I'll have one more PR to come in to update our variant types to account for `Init` and `Empty` states in the `urql` response.

PR #172, opened 2020-05-29
Add Init and Empty constructors to response.

### LINKED ISSUE #144: [Proposal] Init value in response type

parkerziegler:
Fix #144 

This PR adds in two new constructors to our `response` variant in `reason-urql` – `Init` and `Empty`. These address the following conditions and cases in `urql`.

### **`Init`**

Matches the case where a hook has been initialized but no requests have been made for data. This happens in a few cases:

- Using the `useQuery` and `useSubscription` hooks with `pause`. Operations won't be sent until `pause` becomes `true`.
- Using `useMutation` and `useDynamicMutation`. Since mutations don't get executed until the returned `executeMutation` function gets called, we need a state to signal that data hasn't been fetched _yet_.

### **`Empty`**

Matches the case where, for whatever reason, a hook has made a request for data but neither `data` nor `error` have been returned (or are both `null` / `undefined`). This situation happens w/ headless CMSes frequently, especially when you perform a search in your API request (I've seen this with Contentful, for example). `data` is `null` but no errors were returned.

Previously, the two above cases were handled by `NotFound`, which wasn't very descriptive and confused a lot of users. With these two additions, we give more flexibility to users for how to represent different states of their UI.

The one piece I'm unsure of is how we track `hasExecuted` (see diff). I currently use a `ref` to track whether or not the request has been executed – this is what allows us to differentiate between `Init` and `Empty` states. Would love feedback on this piece!


docs/advanced.md
```diff
@@ -0,0 +1,57 @@
+# Advanced
```
parkerziegler, line 1: I began work on some more documentation of these variant types to make them easier to understand and discover. Let me know what you think!

src/hooks/UrqlUseMutation.re
```diff
@@ -34,16 +34,25 @@ let useMutation = (~request) => {
...
+          ~hasExecuted=hasExecuted->React.Ref.current,
+        ),
       (stateJs, parse),
     );
 
   let executeMutation =
     React.useMemo2(
       ((), ~context=?, ()) => {
+        if (!hasExecuted->React.Ref.current) {
+          hasExecuted->React.Ref.setCurrent(true);
```
parkerziegler, line 54: Once we call `executeMutation` for the first time, we set `hasExecuted` to `true`. If we get no data or errors on next render, we'd match the `Empty` constructor.

src/hooks/UrqlUseQuery.re
```diff
@@ -72,16 +72,28 @@ let useQuery =
...
-      () => UrqlResponse.urqlResponseToReason(~response=stateJs, ~parse),
-      (stateJs, parse),
+    React.useMemo3(
+      () =>
+        UrqlResponse.urqlResponseToReason(
+          ~response=stateJs,
+          ~parse,
+          ~hasExecuted=
+            hasExecuted->React.Ref.current
+            || !pause->Belt.Option.getWithDefault(false),
```
parkerziegler, line 85: A little trickier with `useQuery`. We track `hasExecuted` as a ref and also look at the `pause` value. If `pause` is `false` we know the query has already executed, since queries run on mount in this case. If `pause` is `false` then we look at the `hasExecuted` ref.

kitten:
Just a thought before I dive into a review, hopefully by find time tomorrow. We have not opted to add any state charts in the main library since all states are distinct by the presence of every state field on our hooks; so the states are:

- No data, no error, but fetching
- No data, no error, stopped fetching (paused or cancelled)
- Data and no error, not fetching
- Null data, no error, not fetching (technically the `null` case which is distinct from no data, but means we got an empty response)
- Data and error (the presence of errors does not mean there’ll be no data)
- Data and error / Data and no error, and fetching (this is outdated data due to a refetch of any kind)
- Additionally: A `stale` flag can be set to true to indicate that more data is being fetched in the background, which will mean that `fetching` is `false` unless a refetch has changed the query and the state hasn’t reset

So there’s a lot of unrepresented cases in our state diagram / type variants here. I was just wondering how far we want to go 😅 so I’m just throwing them out here. If we go for a simplified   version, that’s fine and intended, but I’ll think about whether there are edge cases where the user would then accidentally obscure state for themselves that’d lead to unexpected UI behaviour

parkerziegler:
Thanks for the comment @kitten. Let me see if I can break this down w/ my read on state -> variant constructor.

- No data, no error, but fetching = `Fetching`
- No data, no error, stopped fetching (paused or cancelled) = `Init` (I'm not familiar with the cancelled case, would `Init` semantically cover that in your mind, or is that a different case).
- Data and no error, not fetching = `Data`
- Null data, no error, not fetching (technically the null case which is distinct from no data, but means we got an empty response)= `Empty`
- Data and error (the presence of errors does not mean there’ll be no data) = good shout on this one, maybe we need `DataAndError(d, e)`?
- Data and error / Data and no error, and fetching (this is outdated data due to a refetch of any kind) = interesting, didn't think about this case. Would `Fetching` work here, or do you think we need another case?

kitten:
I think a complete variant type would be:

```reason
type stateT('data) =
  | Fetching
  | Error(CombinedError.t)
  | PartialResult('data, array(GraphQLError.t))
  | PartialStaleResult('data, array(GraphQLError.t))
  | Result('data)
  | StaleResult('data)
  | StaleEmpty
  | Empty;
```

I'd actually ignore `Init` and `Inactive`; It should sufficiently be covered by the other cases. If I get this right, `Init` is when we'd otherwise use `Empty` but `!fetching` applies? I'd argue that no UI would be interested in that usually, as they'd already be displaying data. This would also only happen when an exchange actively sends a teardown back, which currently only happens for subscriptions in a default environment.

I'd argue that `stale` is important, but we could also encourage people to not use the variant version of the result, if they have to deal with `stale`, partial Graphcache data, or `cache-and-network` updates 🤔 

That brings us to errors and data. This is actually pretty common and I'd consider that most people would either have an API that sends them as metadata (e.g. "the user doesn't have permissions for this field") or as unexpected errors (e.g. TypeError on a deep resolver so the next optional parent is nulled out)

That means that, imho, `Error(NetworkError.t)` would be an interesting case, but the problem is that there can also be "catastrophic" GraphQLErrors, so it must still be `CombinedError.t` there.

So all in all for me that amounts to this:

```reason
type stateT('data) =
  | Fetching
  | Error(CombinedError.t)
  | PartialResult('data, array(GraphQLError.t))
  | Result('data)
  | Empty;
```

Without an `Init` state since it shouldn't usually happen, but an `Empty` state, and a `PartialResult`, where the `Data` is complemented by errors, which could then also circumvent the `CombinedError.t` stuff 🤔  The migration path then would be to use `PartialResult(data, _) | Result(data)` together in a case.

parkerziegler:
Closing this pull request in favor of a new one that I hope to put up this weekend!

Issue #170, opened 2020-05-27
[Umbrella] reason-urql v2 Refactors & Improvements


kitten:
This is a good and opportune moment to take some time to reflect on a couple of `reason-urql`-specific structures and APIs. Specifically, `reason-urql` is meant to be idiomatic Reason code, but some parts of the API have been tied to bindings and expose some of the disconnect between JS/TS and Reason.

Because we're preparing to land v2, this may be a good moment to:

- Improve APIs and make them more idiomatic and friendlier
- Clean up the project structure and reconcile some files
- Update APIs to use Reason idiomatic types as possible.

We can collect ideas in this umbrella issue and close it as soon as `v2` is released.

### Amount of Supporting Files

Currently the project is structured like a JS project with lots of files and some potentially confusing locations.

- The Client has its own folder and is segregated from its types, a separate types file exists
- The exchanges have their own folder with just one module
- Do we still need component support?
- The project's namespacing seems a little aggressive. Can we use BS namespaces better / more effectively?

### Hook Context APIs

A lot of the context is typed and directly accessible. It may make sense to extend the hook API to be more idiomatic for the GraphQL bindings, and to integrate more of the context into the hooks' arguments.

- Should the `request` be its own curried argument?
- Should `unit` be last in those hooks? We do have the required `request` argument.
- `partialOperationContext` could be integrated as a set of labelled optional arguments instead, right into the hook. We could write a small helper for that and pipe all arguments through. `requestPolicy` is duplicate for instance, `url` can be integrated, etc. Optimally the context itself should just be a JSON type and all relevant context options could be surfaced as labelled arguments.

### Other Exchange Bindings

We haven't bound to some crucial exchanges yet and this could be an opportunity to add them?

parkerziegler:
Thanks for opening this @kitten! I'm going to separate each section above into a separate comment with thoughts and a checklist so folks we can discuss each piece by piece.

parkerziegler:
### Amount of Supporting Files

I agree with you, we can certainly clean this up.

**The Client has its own folder and is segregated from its types, a separate types file exists**

I would be happy to move the client just beneath the top-level `src` directory and rejoin the contents of the `ClientTypes` module with it. The main reasons I separated it:

[...]

PR #181, opened 2020-06-06
Flatten folder structure. Make ctx modifications labelled args.

parkerziegler:
This PR makes more progress towards `v2` and addresses a few pieces of feedback from #170. The major improvements:

**Flatter folder structure** 

- Removed `UrqlClientTypes` module altogether (types were moved to either `UrqlClient` or  `UrqlTypes`).
- Moved `UrqlExchanges` into a submodule inside of `UrqlClient`. This was necessary to avoid cyclic dependencies, since `Exchanges` and `Client` both need to reference types from one another.
- Removed the `utils` folder altogether. `UrqlRequest` and `UrqlResponse` both got removed altogether, and their associated functions got moved to files where they're used.

**Context as labelled arguments**

Rather than having a public API where users have to create a `partialOperationContext` and pass that to hooks / `execute` functions, each key on a `partialOperationContext` is now a labelled argument. For example, users can now do the following:

```re
let (state, executeQuery) = useQuery(~request, ());

/* When they want to execute the query. */
executeQuery(~requestPolicy=`NetworkOnly, ~url="https://my-new-graphql-endpoint.com", ());
```

This feels like a much more ergonomic API ✅


__tests__/UrqlResponse_test.re
```diff
@@ -85,64 +85,4 @@ describe("UrqlResponse", () => {
       Expect.(expect(result.response) |> toEqual(UrqlTypes.NotFound));
     });
   });
-
-  describe("urqlClientResponseToReason", () => {
```
parkerziegler, line 89: This function got moved into `UrqlClient` since it's only in use by that module.

package.json
```diff
@@ -28,8 +28,7 @@
   "dependencies": {
     "bs-fetch": "^0.5.2",
     "graphql": "^14.1.1",
-    "urql": "1.5.1",
-    "wonka": "^3.2.2"
```
parkerziegler, line 32: I decided to remove our explicit dependency on `wonka` since it'll be shipped with this version of `urql` and BS can still resolve it just fine. Even though including it appears mostly harmless (it'll get deduped by `yarn`), it just seems like overhead we don't necessarily need. Let me know if I'm wrong here and it's better for us to include it explicitly.

src/UrqlClient.re
```diff
@@ -0,0 +1,380 @@
+type t;
```
parkerziegler, line 1: Just moved this file from `src/client/UrqlClient.re` to `src/UrqlClient.re`. Most other changes are minimal here.

src/hooks/UrqlUseQuery.re
```diff
@@ -1,13 +1,20 @@
...
-  (~context: UrqlClientTypes.partialOperationContext=?, unit) => unit;
+  (
+    ~fetchOptions: Fetch.requestInit=?,
+    ~requestPolicy: UrqlTypes.requestPolicy=?,
+    ~url: string=?,
+    ~meta: UrqlOperations.operationDebugMeta=?,
+    ~pollInterval: int=?,
+    unit
+  ) =>
+  unit;
```
parkerziegler, line 17: A good example of the new API where each key on a `partialOperationContext` is a labelled argument that can be provided to the `execute*` function.

src/UrqlClient.re
```diff
@@ -0,0 +1,373 @@
...
+  switch (fetch) {
+  | Some(FetchWithUrl(impl)) => Some(impl)
+  | Some(FetchWithUrlOptions(impl)) => Some(impl)
+  | Some(FetchWithRequest(impl)) => Some(impl)
+  | Some(FetchWithRequestOptions(impl)) => Some(impl)
+  | None => None
+  };
+};
+
+module Exchanges = {
```
parkerziegler, line 44: The `Exchanges` module is now a submodule of `Client` so each module can safely reference types from the other w/out a cyclic dependency error.

PR #182, opened 2020-06-20
Use BuckleScript's native namespacing feature.

parkerziegler:
This PR adds support for using BuckleScript's native `"namespace: true"` feature. Rather than all of our `module` names being `Urql*` and then renamed in `ReasonUrql.re`, we just use [BuckleScript's native `"namespace: true"`](https://reasonml.org/docs/reason-compiler/latest/build-configuration#name-namespace). This is great – our codebase is easier to traverse and module references are more succint. This to address feedback from #170. Would love to get any feedback or opinions from folks before merging. For the most part, the public API doesn't really change. The one piece that does is `UrqlTypes` is now just `Types`. So if you do the following:

```reason
open ReasonUrql;
```

and you have a local `Types` module, you'd get a conflict. Not a huge deal though, because you can make the (wiser) choice to not globally `open ReasonUrql` and just reference it as needed. If approved, I'll update our code samples to show that pattern a bit more often.


parkerziegler:
Yeah fortunately there's not too much in repo module referencing – mostly just modules referencing `UrqlTypes` and needing to get upgraded to `Types`. But yeah, much smoother than anticipated!

Issue #171, opened 2020-05-28
upgrade urql version


BlueHotDog:
Hey, seems like urql is at 1.8.2 and specifically it seems like some major critical bugs have landed there. 
Anything specific preventing from following up more closely with mainline urql?

parkerziegler:
Hey @BlueHotDog - a few things are blocking this:

- We're in the middle of changing some APIs pretty radically, see #170. We want to have those figured out before continuing to upgrade `urql` versions.
- Each minor version of `urql` introduces changes we can't be sure our bindings support. So in general, we've been conservative by updating the bindings minor version by minor version. If you're really keen to use newer `urql` you could always use yarn resolutions to force a specific `urql` version, but the bindings aren't guaranteed to work 😄 . We'll get there eventually, just doing all of this in my free time so I can't promise super fast upgrades. Feel free to help out with the v2 API improvements listed in that issue linked above if you like, PRs always welcome!

parkerziegler:
Remaining tasks to bring `reason-urql` up to latest `urql` version.

- [ ] Add `operation` field on all hook results.
- [ ] Add `maskTypename` and `preferGetMethod` as options for the `Client`.
- [ ] Add `client.readQuery` method. @tatchi has a PR open for this now that we need to re-point at `main`.
- [ ] Add utilities for `makeResult`, `makeErrorResult`, `formatDocument`, and `maskTypename`.
- [ ] Add `enableAllOperations` option on the `subscriptionExchange`.
- [ ] Add `additionalTypenames` field on the operation context record type.
- [ ] Add `client.subscription` method.

Folks should feel free to take any or all of the above – just comment on this thread with which binding you're working on!

parkerziegler:
Latest `reason-urql` release is bound to `urql` v1.10.0!

PR #198, opened 2020-08-05
Migrate to urql v1.7.0.

parkerziegler:
This PR brings us up to `urql` v1.7.0! Most of the code included this PR are small updates to our examples, but no major changes to the bindings. Based on [the CHANGELOG for `urql` v1.7.0](https://github.com/FormidableLabs/urql/blob/v1.7.0/CHANGELOG.md#v170), it seemed like the major change of that version was the splitting of `urql` into separate entry points.

I'm not fully sure how we want to handle that here. From my perspective, the majority use case for using `reason-urql` is in conjunction with ReasonReact, so I'm not yet sold on splitting the bindings into a separate entry point as well (though I would love other opinions and discussion on this). If we did decide to support the separate entry point, I think it'd mostly involve:

- Modifying our `[@bs.module]` bindings for certain parts of the API to point at `@urql/core` rather than `urql`.
- Potentially mirroring the setup of `urql` with yarn workspaces and `@reason-urql/core` and `reason-urql` as two separate workspaces.

I'm not sure what we think, so let's chat!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,10 @@ type t =
 - Component bindings for `Query`, `Mutation`, `Subscription`, and `SubscriptionWithHandler` were removed. Just use the hooks APIs!
 - `Client` methods for `executeRequestOperation`, `reexecuteOperation`, `createRequestOperation`, and `dispatchOperation` were removed. These are no longer in `urql`'s public facing API.
 
+### Diff
+
+https://github.com/FormidableLabs/reason-urql/compare/v1.7.0...v2.0.0
+
 ## [1.7.0] - 2020-04-16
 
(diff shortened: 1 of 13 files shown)


kitten:
Believe it or not; I think you can go all the way to `urql@1.9.8` without any breaking changes or API changes, since a lot of patches were spent on getting concurrent-related timings right.

This will also skip over some versions that should likely not be used anymore as they just don't always behave correctly, either because of Suspense or Concurrent Mode.

For `urql@1.10.0` you'd only need the `operation` field on all hook results.

For `@urql/core` we've got:

- The addition of `maskTypename` as a utility and option
- The `client.readQuery` method
- The `enableAllOperations` option on the `subscriptionExchange`
- The `additionalTypenames` field on the Operation context

That brings you up to `@urql/core@1.11.8`. For `v1.12.3` you'd then need to add the `client.subscription` shortcut method.

Great work! We're not far off!

parkerziegler:
Sounds great! Merging this for now, and I'm going to move the above points into #171 to track 🚀

PR #208, opened 2020-09-04
Bind last APIs available in urql 1.10.

parkerziegler:
This PR binds the final APIs from `urql` 1.10 that I have slated for addition in #171. These include:

- Add utilities for makeResult, makeErrorResult, formatDocument, and maskTypename.
- Add enableAllOperations option on the subscriptionExchange.
- Add `client.subscription` method.

There was also a change that involved removal of our final `[@bs.deriving abstract]`, `partialOperationContext`. From what I can tell in the new ReScript docs, this is pretty much going away entirely in ReScript, so I decided to plan ahead for it. The big bummer about direct record -> object compilation, however, is that it doesn't really provide a good mechanism for generating `Partial` objects like in TS. We can't define half of these keys on a record, for example – if they're potentially undefined we have to explicitly set their value to `None`. This doesn't translate cleanly to the object spread syntax used in `urql`, because now ReScript will generate a `context` object like:

```js
{
  fetch: undefined,
  fetchOptions: undefined,
  url: undefined,
  requestPolicy: undefined
  // etc.
}
```

that'll overwrite the configuration pulled in by the Client. I'll point these places out in the code if anyone has any solid ideas. From what I can tell, partial object compilation isn't really possible without `[@bs.deriving abstract]`, which feels like the tradeoff for explicitness.

### DIFF
--- a/examples/1-execute-query-mutation/webpack.config.js
+++ b/examples/1-execute-query-mutation/webpack.config.js
@@ -17,7 +17,7 @@ module.exports = {
     alias: {
       react: path.resolve("./node_modules/react"),
       "react-dom": path.resolve("./node_modules/react-dom"),
-      urql: path.resolve("./node_modules/urql"),
+      urql: path.resolve("../../node_modules/urql"),
     },
   },
 };
--- a/examples/2-query/src/index.re
+++ b/examples/2-query/src/index.re
@@ -1,7 +1,7 @@
 open ReasonUrql;
 
 /* Instatiate the client. */
-let client = Client.make(~url="https://graphql-pokemon.now.sh", ());
+let client = Client.make(~url="https://graphql-pokemon2.vercel.app", ());
 
 ReactDOMRe.renderToElementWithId(
   <Context.Provider value=client> <Container /> </Context.Provider>,
--- a/src/Client.re
+++ b/src/Client.re
@@ -1,4 +1,10 @@
-type t;
+type t = {
+  url: string,
+  suspense: bool,
+  preferGetMethod: bool,
+  requestPolicy: string,
+  maskTypename: bool,
+};
 
 /* Helpers for supporting polymorphic fetchOptions. */
 type fetchOptions('a) =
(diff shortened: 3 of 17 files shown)


examples/2-query/src/index.re
```diff
@@ -1,7 +1,7 @@
 open ReasonUrql;
 
 /* Instatiate the client. */
-let client = Client.make(~url="https://graphql-pokemon.now.sh", ());
+let client = Client.make(~url="https://graphql-pokemon2.vercel.app", ());
```
parkerziegler, line 4: Thew Now 1.0 deployment got axed, but fortunately they put up a new deployment 🎉

examples/1-execute-query-mutation/webpack.config.js
```diff
@@ -17,7 +17,7 @@ module.exports = {
     alias: {
       react: path.resolve("./node_modules/react"),
       "react-dom": path.resolve("./node_modules/react-dom"),
-      urql: path.resolve("./node_modules/urql"),
+      urql: path.resolve("../../node_modules/urql"),
```
parkerziegler, line 20: This wasn't being consistently resolved in examples using symlinking, so I decided to always resolve at root `node_modules`.

src/Client.re
```diff
@@ -1,4 +1,10 @@
-type t;
+type t = {
+  url: string,
+  suspense: bool,
+  preferGetMethod: bool,
+  requestPolicy: string,
+  maskTypename: bool,
+};
```
parkerziegler, line 7: We only expose _some_ properties on `Client` publicly (not the methods). `url` is the one that I can see being most commonly sought, and we use it internally to generate safe fallback urls.

src/Client.re
```diff
@@ -234,21 +241,23 @@ let executeQuery =
...
+      ~variables=request##variables,
       (),
     );
+  let parse = request##parse;
+  let optsJs: Types.partialOperationContext = {
+    additionalTypenames,
+    fetchOptions,
+    fetch,
+    requestPolicy: requestPolicy->Belt.Option.map(Types.requestPolicyToJs),
+    url: Some(url->Belt.Option.getWithDefault(client.url)),
```
parkerziegler, line 255: This makes me 😭 With our labeled argument being passed into this function, we allow users to overwrite the `url` that `executeQuery` requests. This is good! However, when composing this object, if a user _doesn't_ supply the `url` parameter, the `context` object will get compiled to have `url: undefined`. This then overwrites the `url` supplied by the Client, since `context` is spread into `executeQuery`. So we ensure here that we safely fallback to `client.url` if no `url` is provided. Annoying result of ReScript not really having good partial object support.

src/hooks/UseQuery.re
```diff
@@ -56,25 +64,50 @@ let useQuery =
...
-      ~url?,
-      ~pollInterval?,
-      ~meta?,
-      ~suspense?,
-      ~preferGetMethod?,
-      (),
+
+  let client = UseClient.useClient();
+
+  let context =
```
parkerziegler, line 74: So this is a big bummer. Without this memoization, we create a new `context` object on every render, which causes `useQuery` to run, which re-renders, which generates a new `context` object, and boom, we're off to infinite re-renders. So we have to memoize the `context` object. Unfortunately `reason-react` only provides `useMemo` arity up to 7 so I had to bring in `useMemo10` 😂

PR #213, opened 2020-09-08
Improve handling of potential null values from the JS side.

### LINKED ISSUE #202: Runtime error "value is null" with the latest Reason Urql version
I get the following runtime error when trying to update to Reason Urql 2:
Capture d’écran 2020-08-19 à 11 43 28

The `parse` function is generated by the `graphql ppx` and will indeed accept an object, not null.

After inverstigating a bit more, I could find that the following function (which I use), https://github.com/FormidableLabs/reason-urql/blob/main/src/Client.re#L221 will pass the `parse` function to https://github.com/FormidableLabs/reason-urql/blob/main/src/Client.re#L188, which in turn, will use the `Belt.Option.map` function to conditionally call the `parse` function.

The issue arises because of the type of `response`, which is assumed to be https://github.com/FormidableLabs/reason-urql/blob/main/src/Types.re#L81, where `data` is an `option(Js.Json.t)`, which is not _at runtime_, it's actually a `Js.Nullable.t(Js.Json.t)`.

But maybe I'm doing something wrong?

_Edit_: The error attribute is also "wrong" since it's assumed to be an option as well. Also, it seems this function https://github.com/FormidableLabs/reason-urql/blob/main/src/Types.re#L135 has the same issue 😕 

_Edit 2_: With the `-lean-parse` option turned off, I get this error instead:
Capture d’écran 2020-08-19 à 12 38 14

parkerziegler:
This PR attempts to fix #202

@gaku-sei I'd love your 👀 here. The main gist of the changes:

- Treat all fields on `operationResult` as `Js.Nullable.t`s except for `operation` (which is guaranteed to be defined).
- Ensure we convert `Js.Nullable` values like `data`, `error`, `extensions`, and `stale` to `option` before operating on them.

I read your comment in #202 and agree that it's a bit uncertain where we really need the `Js.Nullable` value. From my perspective, I know that `urql` generally does a good job avoiding passing `null` in its internals, so direct `option` compilation to handle `undefined` feels ok for most cases. But agree that anything being passed through from the GraphQL server should be treated with a bit more caution.

### DIFF
--- a/src/Client.re
+++ b/src/Client.re
@@ -190,7 +190,9 @@ type response('response) =
 type clientResponse('response) = {
   data: option('response),
   error: option(CombinedError.t),
+  extensions: option(Js.Dict.t(string)),
   response: response('response),
+  stale: option(bool),
 };
 
 /**
@@ -198,9 +200,13 @@ type clientResponse('response) = {
  * methods to typed a Reason record.
  */
 let urqlClientResponseToReason = (~response: Types.operationResult, ~parse) => {
-  let data = response.data->Belt.Option.map(parse);
+  let data = response.data->Js.Nullable.toOption->Belt.Option.map(parse);
   let error =
-    response.error->Belt.Option.map(CombinedError.combinedErrorToRecord);
+    response.error
+    ->Js.Nullable.toOption
+    ->Belt.Option.map(CombinedError.combinedErrorToRecord);
+  let extensions = response.extensions->Js.Nullable.toOption;
+  let stale = response.stale->Js.Nullable.toOption;
 
   let response =
     switch (data, error) {
@@ -209,7 +215,7 @@ let urqlClientResponseToReason = (~response: Types.operationResult, ~parse) => {
     | (None, None) => Empty
     };
 
-  {data, error, response};
+  {data, error, extensions, stale, response};
 };
 
 /* Execution methods on the client. These allow you to imperatively execute GraphQL
--- a/src/Client.re
+++ b/src/Client.re
@@ -200,13 +200,11 @@ type clientResponse('response) = {
  * methods to typed a Reason record.
  */
 let urqlClientResponseToReason = (~response: Types.operationResult, ~parse) => {
+  let {extensions, stale}: Types.operationResult = response;
+
   let data = response.data->Js.Nullable.toOption->Belt.Option.map(parse);
   let error =
-    response.error
-    ->Js.Nullable.toOption
-    ->Belt.Option.map(CombinedError.combinedErrorToRecord);
-  let extensions = response.extensions->Js.Nullable.toOption;
-  let stale = response.stale->Js.Nullable.toOption;
+    response.error->Belt.Option.map(CombinedError.combinedErrorToRecord);
 
   let response =
     switch (data, error) {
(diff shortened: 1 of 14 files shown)


src/Client.re
```diff
@@ -190,17 +190,23 @@ type response('response) =
 type clientResponse('response) = {
   data: option('response),
   error: option(CombinedError.t),
+  extensions: option(Js.Dict.t(string)),
```
parkerziegler, line 193: We were missing `extensions` 😱 Unfortunately the `urql` type for this field, `Record<string, any>` is hard to bind without introducing _lots_ of type parameters. At the same time, `Js.Dict.t(string)` limits extensions to being JS objects with only string values 😞 I'd like to create a better solution down the road.

src/Client.rei
```diff
@@ -133,7 +133,9 @@ type response('response) =
 type clientResponse('response) = {
   data: option('response),
   error: option(CombinedError.t),
+  extensions: option(Js.Dict.t(string)),
   response: response('response),
+  stale: option(bool),
```
kitten, line 138: My memory sometimes doesn't serve me too well but I believe this field is guaranteed

parkerziegler: So it _is_ guaranteed on the responses handed to hooks, but not on `OperationResult`, which is what gets returned by Client.* methods. So in this case we pass it through with the `option` wrapper and let folks decode it if they like.

src/Client.re
gaku-sei: Indeed, sorry, I forgot to report this ^^

An alternative could be an opaque type maybe?

```reasonml
type extension;

extensions: option(Js.Dict.t(extension))
```

If `any` in urql stands for "any _json_ value", then `extensions: option(Js.Json.t)` should work?

gaku-sei:
Thank you @parkerziegler 🙏 

To be honest I don't know the js/ts api very well, so I was bit stuck 😕 

Overall looks really good, and after testing on my "update to v2" branch, it seems to work (i.e. no more runtime errors 🎊 ). I just have an other small issue with https://github.com/FormidableLabs/reason-urql/issues/215 now 😄 

TS <-> Reasonml typings can be really hard, do you thing https://github.com/reason-association/genType could help a bit?

src/Client.re
parkerziegler: Well `extensions` in the GraphQL spec are a well-ish defined structured. Per the GraphQL spec:

> The response map may also contain an entry with key extensions. This entry, if set, must have a map as its value.

https://spec.graphql.org/June2018/#sec-Response-Format

So we know `extensions` will always be a map (read: object) structure in JS,  but the types of the values _may_ be polymorphic. In the vast majority of cases I'd expect all values to be `strings`, but this isn't really a safe assumption.

I do like your idea of using `Js.Json.t` and then allowing users to decode it. Their API can set the shape of extensions and they can write their own decoders using the `Js.Json` module 👍

parkerziegler:
I don't think `genType` will be too much help here. We'd be wanting to `import` the types from `urql` to see what sorts of types they generate in Reason, but from what I can tell `genType` will always leave these types opaque: https://github.com/reason-association/genType#export-and-import-types

parkerziegler:
@gaku-sei I merged this and am going to merge #216 as well and put out an rc release for testing!

PR #216, opened 2020-09-09
Revert to using bs.deriving abstract for partialOperationContext.

### LINKED ISSUE #215: Client.executeQuery will "reset" all the default values of client
When using the `Client.executeQuery` function, it seems all the default values are dropped (fetchOptions, fetch, etc...).

It seems related to how the options are defined [here](https://github.com/FormidableLabs/reason-urql/blob/88e8e0418e304924812d12bab49f09d2df087594/src/Client.re#L250). This object will be passed down to the `executeQuery` function which will spread it [here](https://github.com/FormidableLabs/urql/blob/main/packages/core/src/client.ts#L176).

The issue is that the previous properties will be replaced by the new ones, with value "undefined":

```reason
Client.executeQuery(~client, ~request, ());

// Will generate an opts as follow:

let optsJs: Types.partialOperationContext = {
  additionalTypenames,
  fetchOptions: undefined, // since the type is option('fetchOptions),
  fetch: undefined, // since the type is option('fetchImpl),
  // ...
};
```

When spread in the `createOperationContext` method:

```ts
{
  url: this.url,
  fetchOptions: this.fetchOptions,
  fetch: this.fetch,
  // ...
  ...opts,
  // ...
};
```

Which is the same as

```ts
{
  url: this.url,
  fetchOptions: this.fetchOptions,
  fetch: this.fetch,
  // ...
  fetchOptions: undefined,
  fetch: undefined,
  // ...
};
```

So

```ts
{
  url: this.url,
  // ...
  fetchOptions: undefined,
  fetch: undefined,
  // ...
};
```

Since the properties are spread _after_ the default values (which is necessary to allow for default values override).

A solution is to make the [clientOptions](https://github.com/FormidableLabs/reason-urql/blob/88e8e0418e304924812d12bab49f09d2df087594/src/Client.re#L141) type abstract (just like `PartialOperationContextJs`).

But I'm too sure about the urql api, so maybe I missed something? 😅

parkerziegler:
Fix #215 

This PR ensures that `partialOperationContext` always returns a truly "partial" object. Using direct object -> record compilation is problematic for this case, because `None` will compile to `undefined` and wipe out defaults if the compiled object is being used with object spread syntax.

@gaku-sei this should fix your issues, because the compiled context objects will only include keys for properties that are passed in the labeled arguments for `executeQuery` as `Some(value)`.

### DIFF
--- a/src/Client.re
+++ b/src/Client.re
@@ -247,17 +247,19 @@ let executeQuery =
       (),
     );
   let parse = request##parse;
-  let optsJs: Types.partialOperationContext = {
-    additionalTypenames,
-    fetchOptions,
-    fetch,
-    requestPolicy: requestPolicy->Belt.Option.map(Types.requestPolicyToJs),
-    url: Some(url->Belt.Option.getWithDefault(client.url)),
-    pollInterval,
-    meta,
-    suspense,
-    preferGetMethod,
-  };
+  let optsJs =
+    Types.partialOperationContext(
+      ~additionalTypenames?,
+      ~fetchOptions?,
+      ~fetch?,
+      ~requestPolicy=?requestPolicy->Belt.Option.map(Types.requestPolicyToJs),
+      ~url?,
+      ~pollInterval?,
+      ~meta?,
+      ~suspense?,
+      ~preferGetMethod?,
+      (),
+    );
 
   executeQueryJs(~client, ~query=req, ~opts=optsJs, ())
   |> Wonka.map((. response) => urqlClientResponseToReason(~response, ~parse));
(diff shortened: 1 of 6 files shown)


gaku-sei:
Woah! Thank you very much @parkerziegler 😄 

It seems everything works perfectly now! I will try to keep you updated if there is any regression, but it seems 2.1.0-rc.0 solved all the issues 🎊

parkerziegler:
Awesome keep me updated. Also, thanks for all your help finding and proposing fixes for these issues @gaku-sei, going to add you to the All Contributors list!

@all-contributors please add @gaku-sei for code, bug, ideas.

parkerziegler:
@all-contributors please add @gaku-sei for code, bug, ideas.

Issue #220, opened 2020-09-27
urql shouldn't be a dep but a peer dep


MoOx:
Hi again,

I will make this short. I am still trying to use reason-urql with next!
I am getting this issue:

<img width="987" alt="image" src="https://user-images.githubusercontent.com/157534/94361518-1b2dd080-00b5-11eb-88c1-6938fbd56026.png">

As you can see, the code produced by reason-urql seems zero-cost at some point (I can tell when I read `[@bs.module "urql"] external`). But when you do such bindings, you need to be sure that the dep in question (urql here) is available at the correct level in node_modules. Here my `./pages/index.bs.js` try (because of the code produced by rescript/bucklescript) to reach urql itself. And it's not available since it's specified as a dep in reason-urql.

For reason-react-native for example (which are only zero cost bindings), react-native is a peer dep & not a dep.

You should probably remove urql from deps & put it as peer dep + document this in the installation section of the readme.

parkerziegler:
Hey @MoOx, I began looking but have had a recent death in my family and won't be getting back to it for a few days (or, more likely, a week and a half). My first impressions:

- As far as I know, `urql` should still be located at `/node_modules/urql` after installing `reason-urql`. I suppose I can't say for sure on `npm`, but a fresh installation using `yarn` does yield this result.

<img width="292" alt="image" src="https://user-images.githubusercontent.com/19421190/94850202-d4591700-03e3-11eb-9fae-0614bd1f53b7.png">

So I'm not sure that the location of `urql` in `node_modules` is the root problem here. In general, I believe package managers since npm@3 flatten all dependencies of dependencies, so whether `urql` is a `peerDep` or a `dep` it would still be located at `/node_modules/urql`.

Are you doing anything in `next.config.js` that would alter the standard Node resolution algorithm? As far as I understand Node resolution, it'll start at the root and recursively move one directory up searching for `node_modules`, until it reaches globally installed dependencies.

Whether or not `urql` should be a direct or peer dep is a complicated question because of `urql`'s versioning. If `urql` wasn't a direct dependency it would be very difficult to ensure that the version of bindings a user was using (`reason-urql`) was compatible with their `urql` version. New `urql` versions tend to be released ahead of `reason-urql`, and we don't have sound guarantees that our bindings will work with _any_ minor version.

@kitten I'd be curious about your thoughts here. We did have `urql` as a `peerDependency` in the past but opted to make it a direct dependency to support:

- Single package install
[...]

Issue #219, opened 2020-09-22
Adopt `graphql-ppx` 1.0


jfrolich:
Are you looking to upgrade to `graphql-ppx` 1.0? It's a bit of a change but looking at the examples it allows to refactor the API to the following:

```reason
let request = GetPokemon.make(~name=pokemon, ());
let (Hooks.{response}, _) = Hooks.useQuery(~request, ());
```

into

```reason
let (response, _) = GetPokemon.use({name: pokemon});
```

Also you get all the other goodies like many bugfixes, better error messages, and - most importantly - records instead of objects.

You can extend queries/mutations/subscriptions, and this allows you to add the `use` function (see https://graphql-ppx.com) to create quite a nice and concise API surface. If you have any questions I am happy to help out!

BlueHotDog:
Yes Please!! super looking forward to this!

parkerziegler:
Hey @jfrolich! I'd definitely be interested in adopting it. If you'd like to start working on it, the contribution is more than welcome! I'll likely be swamped w/ work until this week, so that'd be the earliest I could get to it. But it's good to see that a community standard has emerged and yes, I'd be happy to officially move onto it 🎉

jfrolich:
I don't use `reason-urql` but happy to answer questions!

BlueHotDog:
I'm now giving it a try on a project. will see the exceptions i get and maybe i can help push this forward.

parkerziegler:
Hey folks, I began working on this migration but have had a recent death in my family and won't be getting back to it for a few days (or, more likely, a week and a half). My first impressions:

- This looks like a great library! It does seem that it really encourages using functors as the main mechanism for accessing `query`, `parse`, etc. which are the main internals `reason-urql` cares about. However, many of our bindings and APIs as they stand aren't really setup for functors, and I've seen a healthy amount of reports from community members in the past that functors are a serious point of confusion for them. As it stands, I don't think we'd be able to easily adopt this library without a massive overhaul of our APIs, which I'm reticent to do.
- I do think there's a way around this, however (and maybe it's already possible and I'm just not quite getting it). I'd love to be able to access _everything_ off of `module type Operation` _without_ using functors, so that we have all internals of an `Operation` module available to our converter functions. In `graphql_ppx_re` this was available as the `Js.t` returned by calling `.make`, i.e.

```reason
let request = MyQuery.make();

let parse = request##parse;
let query = request##query;
let variables = request##variables;
```

[...]

Issue #206, opened 2020-08-27
TypeError: state.talkback is not a function


MoOx:
Hi, I am having issue with next & getStaticProps. I am getting a weird wonka runtime error.
Here is a sample of my code (still WIP)

```reason
let getStaticProps: Next.GetStaticProps.t(props, {.}) =
  _context => {
    let request =
      GetItems.make(
        ~first=perPage,
        ~pageSlug="noop",
        ~categorySlug="",
        // ~tagSlug=None,
        // ~cursorAfter=None,
        (),
      );

    Client.query(~client=Graphql.client, ~request, ())
    |> Js.Promise.then_(data => {
         switch (Client.(data.response)) {
         | Data(d) =>
           {"props": makeProps(~items=Js.Nullable.return(d), ())}
           ->Js.Promise.resolve
         | Error(e) =>
           {"props": makeProps(~items=Js.Nullable.null, ())}
           ->Js.Promise.resolve
         | Empty =>
           {"props": makeProps(~items=Js.Nullable.null, ())}
           ->Js.Promise.resolve
         }
       })
    |> Js.Promise.catch(err => {
         Js.log2("Failure!!", err);
         {"props": makeProps(~items=Js.Nullable.null, ())}
         ->Js.Promise.resolve;
       });
  };
```

Here is the runtime error I got

```
Failure!! TypeError: state.talkback is not a function
    at _0 (webpack-internal:///./node_modules/wonka/src/operators/wonka_operator_take.js:62:28)
    at eval (webpack-internal:///./node_modules/wonka/src/operators/wonka_operator_takeLast.js:18:18)
    at eval (webpack-internal:///./node_modules/wonka/src/wonka_helpers.js:21:12)
    at eval (webpack-internal:///./node_modules/wonka/src/operators/wonka_operator_take.js:50:14)
    at Module._1 (webpack-internal:///./node_modules/bs-platform/lib/es6/curry.js:55:12)
    at Module.captureTalkback (webpack-internal:///./node_modules/wonka/src/wonka_helpers.js:16:73)
    at eval (webpack-internal:///./node_modules/wonka/src/operators/wonka_operator_takeLast.js:10:79)
    at Module._1 (webpack-internal:///./node_modules/bs-platform/lib/es6/curry.js:55:12)
    at eval (webpack-internal:///./node_modules/wonka/src/web/wonka_sink_toPromise.js:10:75)
    at new Promise (<anonymous>)
```

I have no clue on how to fix this problem...

[...]

PR #221, opened 2020-10-12
Begin migration to @reasonml-community/graphql-ppx, starting with client.execute* methods.

parkerziegler:
This PR beings to address #219. Thanks to @jfrolich and @jeddeloh for their ideas and help here 🙌 

## Summary

Now that the Reason community is beginning to coalesce around [a standard ](https://github.com/reasonml-community/graphql-ppx)(from what I can tell) for a GraphQL PPX rewriter, we want to ensure `reason-urql` is compatible with that standard. Because this process will lead to breaking changes, I'm planning to roll it into a `v3` release alongside fixes for #220 and #206.

### Major Changes

The major breaking changes in this particular PR involve those related to working with the `graphql-ppx` API. Fortunately the changes aren't _that_ wild. Previously we'd do:

```reason
module LikeDog = [%graphql
  {|
  mutation likeDog($key: ID!) {
    likeDog(key: $key) {
      name
      breed
      likes
    }
  }
|}
];

let request = LikeDog.make(~key="12345", ());

Client.executeMutation(~request, ())
  |> Wonka.subscribe((. result) => {
    /* Handle result, error, etc. */
  });
``` 

The new API uses a feature of OCaml called [first-class modules](https://dev.realworldocaml.org/first-class-modules.html) to allows us to use a `graphql-ppx` module almost like a regular variable. The new API looks like this:

```reason
module LikeDog = [%graphql
  {|
  mutation likeDog($key: ID!) {
    likeDog(key: $key) {
      name
      breed
      likes
    }
  }
|}
];

Client.executeMutation(~request=(module LikeDog), ~variables=LikeDog.makeVariables(~key="12345"), ())
  |> Wonka.subscribe((. result) => {
    /* Handle result, error, etc. */
  });
```

I like this quite a bit more! It would allow us to match `urql` naming a bit more closely as well by changing `~request` to `~query` if we like.

### Drawbacks

I'm not sure if you'd necessarily call this a drawback, but this change does make the source less approachable for newcomers. Supporting `graphql-ppx` in this way requires the use of GADTs. Succinctly, GADTs allows us to specify the type of the _output_ of a function based on the type of the _input_ – a great way to handle polymorphism! However, they do take a bit of work to understand. Similarly, first-class modules require some ramping up on to get comfortable with (though I do think they are more approachable than functors)!

### Next Steps

After this PR gets merged, we can begin moving the hooks APIs over to the use of first-class modules. When that is complete, we'll migrate `urql` to a peerDependency and complete the last steps of v3!


jfrolich:
Nice!

With the `extendQuery` configuration in `bsconfig.json` you could simplify to:

```reason
module LikeDog = [%graphql
  {|
  mutation likeDog($key: ID!) {
    likeDog(key: $key) {
      name
      breed
      likes
    }
  }
|}
];

LikeDog.execute({key: "12345"})
|> Wonka.subscribe((. result) => {
    /* Handle result, error, etc. */
  });

// or even
LikeDog.subscribe({key: "12345"}, (. result) => {
  /* Handle result, error, etc. */
})
```

Using applied functors under the hood in the generated code but the user doesn't have to know :)

jfrolich:
I don't understand where GATD's are used?

BlueHotDog:
SOOOO excited for this landing!!!

parkerziegler:
@jfrolich GADTs are used here: https://github.com/FormidableLabs/reason-urql/blob/26fad3868de7e89d993fe4f28754514788a8e53e/src/Client.re#L234 It's opaque to the user, so the drawback I discuss is more for contributors – for users the experience is better from my perspective.

Since we need to correlate the type of `Operation.t` to the return type of the data after it's been parsed by `clientResponseToReason`, we need GADTs to help handle that polymorphism. I took this approach from `reason-apollo-client`, which does a very similar trick (and is a nice solution for handling polymorphism introduced by every GraphQL query / mutation having a different shape).

I think the `LikeDog.execute` API is a more controversial choice for us, and one I'll want to discuss with the `urql` core team. A major motivation we have for this project is to keep the APIs as similar as possible while also taking advantage of the benefits of Reason to improve the ergonomics. But the farther we get away from the core `urql` API the harder it is for us to keep releases synced and ensure users feel confident switching between the two. Hope that makes sense!

jfrolich:
Hey, I think the name for that is locally abstract type. I wrote the early iteration of that heavily inspired by @fakenickels.

Totally understand the intention of staying close to the JS syntax. Just wanted to make sure you didn't miss this option.

parkerziegler:
Ah gotcha, thanks @jfrolich! Definitely will also keep the API extensions in mind as `v3` develops here – it may be that we move towards that API-style a bit more, or provide an opt-in for users.

PR #224, opened 2020-10-30
Make urql a peerDependency. Update examples to use pure CSS.

parkerziegler:
This PR continues on work for releasing `v3`. This should hopefully begin to address #220. `urql` will be a `peerDependency` going forward, with the `peerDependency` range kept fairly strict (to patch upgrades only). The goal here is to keep `reason-urql` bindings more in line with latest `urql` releases. Our examples now validate against `urql` v1.10.3 (released earlier today), even if we may not support the full API (at least not all exchanges).

It isn't clear to me from my time in the Reason community how bindings to libraries should treat their relationship to the base library they bind (direct vs. peer dependency). `reason-react` is a good example of a project that explicitly specifies the libraries it binds to (`react` and `react-dom`) as peer dependencies. That used to not be safe for us because we were so far behind core `urql`. But now that we're in parity, more or less, it feels like a safer choice. This should also help with dependency resolution decisions made by the ReScript compiler when we jump to latest `bs-platform` (v8).

The only other major change in this PR was the dropping of `bs-css` and `bs-css-emotion` in our examples. They added a lot of bloat and extra code for something CSS does beautifully already. I also want to keep our examples much more focused on just _our_ library and not introduce any extra tools for readers to grok.

@JoviDeCroock @kitten I don't think it's worth much time reviewing this one – lots of the code is just examples cleanup. The core change is really just making `urql` a peerDep. Let me know thoughts on that decision. If we don't feel good about it, this work won't all be lost – these examples were getting a little stale and needed an upgrade!


package.json
```diff
@@ -28,25 +28,28 @@
...
     "lint-staged": "^10.2.2",
     "npm-run-all": "^4.1.5",
-    "reason-react": "^0.7.0"
+    "reason-react": "^0.7.0",
+    "urql": "~1.10.1"
   },
   "peerDependencies": {
     "@reasonml-community/graphql-ppx": "^1.0.1",
-    "reason-react": "^0.7.0"
+    "graphql": "^15.3.0",
```
kitten, line 50: I've only seen this now; I believe the GraphQL PPX doesn't have a peer deps envy range for GraphQ. Given the package's little changes this will need to probably be as loose as urql's range on GraphQL: `^0.11.0 || ^0.12.0 || ^0.13.0 || ^14.0.0 || ^15.0.0` (although you obviously don't have to go as low)

package.json
```diff
@@ -28,25 +28,28 @@
...
+    "reason-react": "^0.7.0",
+    "urql": "~1.10.1"
```
kitten, line 52: I think it's fine to allow minor ranges now? I get the motivation but given that all `~1.10.1` patches have been quite important and the ones following have been as well we could go to `^1.13.0` or `^1.14.0` which don't actually have any relevant changes

kitten (approved):
This is so much 😅❤️ I didn't quite go through the code changes of all examples but omg I'm so excited 🙌

package.json
parkerziegler: Ah good catch, will update!

package.json
parkerziegler: So as I understand it, this `peerDependency` is really on `react-urql` rather than `@urql/core`. `react-urql` (from what I can tell by the `package.json` here: https://github.com/FormidableLabs/urql/blob/main/packages/react-urql/package.json#L3 is on `v1.10.3`. That's what trying to bind to w/ this range (which looks like it'll bring along `@urql/core@1.14.1` 🙌

JoviDeCroock (approved):
Woo, this must've required a lot of work 😅 looks great Parker

kitten:
Quick note 😅 was the peer on react-Dom accidental or is there something I'm missing here?

parkerziegler:
@kitten Woops accidental! I'm so used to adding them together that I just autopiloted for a second. Will amend in a bit!

PR #225, opened 2020-11-01
Upgrade hooks api to use graphql-ppx 1.0

amiralies:
* [X] useQuery
* [X] useMutation
* [x] useSubscription
* [x] query examples
* [x] mutation examples
* [x] subscription examples

### DIFF
--- a/examples/2-query/public/index.css
+++ b/examples/2-query/public/index.css
@@ -13,6 +13,7 @@
   border-radius: 5px;
   box-shadow: 0 10px 15px -3px rgba(0, 0, 0, 0.1),
     0 4px 6px -2px rgba(0, 0, 0, 0.05);
+  margin-top: 15rem;
 }
 
 .dex > * + * {
--- a/examples/2-query/public/index.css
+++ b/examples/2-query/public/index.css
@@ -1,7 +1,4 @@
 .dex-container {
-  display: flex;
-  align-items: center;
-  justify-content: center;
   margin: auto;
 }
 
@@ -13,7 +10,6 @@
   border-radius: 5px;
   box-shadow: 0 10px 15px -3px rgba(0, 0, 0, 0.1),
     0 4px 6px -2px rgba(0, 0, 0, 0.05);
-  margin-top: 15rem;
 }
 
 .dex > * + * {
@@ -60,15 +56,14 @@
 }
 
 .sidebar-container {
-  position: absolute;
+  display: flex;
   inset: 0;
-  display: grid;
-  grid-template-columns: minmax(20rem, 20%) 1fr;
 }
 
 .sidebar {
   display: flex;
   flex-direction: column;
+  height: 100vh;
   background: #f25050;
   padding: 1rem;
   overflow: auto;
--- a/examples/2-query/src/Container.re
+++ b/examples/2-query/src/Container.re
@@ -27,7 +27,7 @@ let make = () => {
     Hooks.useQuery(
       ~query=(module GetAllPokemons),
       ~requestPolicy=`CacheFirst,
-      GetAllPokemons.makeVariables(~first=151, ()),
+      {first: 151},
     );
 
   switch (response) {
--- a/examples/2-query/src/Container.re
+++ b/examples/2-query/src/Container.re
@@ -11,32 +11,35 @@ module GetAllPokemons = [%graphql
 |}
 ];
 
-let request = GetAllPokemons.make(~first=151, ());
-
-let flattenPokemon = pokemons =>
-  pokemons
-  |> Array.map(pokemon =>
-       switch (pokemon) {
-       | Some(pokemon) => Belt.Option.getWithDefault(pokemon##name, "")
-       | None => ""
-       }
-     );
+let flattenPokemon = pokemons => {
+  GetAllPokemons.(
+    pokemons->Belt.Array.map(pokemon =>
+      pokemon
+      ->Belt.Option.flatMap(p => p.name)
+      ->Belt.Option.getWithDefault("")
+    )
+  );
+};
 
 [@react.component]
 let make = () => {
   let (Hooks.{response}, _) =
-    Hooks.useQuery(~request, ~requestPolicy=`CacheFirst, ());
+    Hooks.useQuery(
+      ~query=(module GetAllPokemons),
+      ~requestPolicy=`CacheFirst,
+      GetAllPokemons.makeVariables(~first=151, ()),
+    );
 
   switch (response) {
   | Fetching => <div> "Loading"->React.string </div>
   | Data(data) =>
-    switch (data##pokemons) {
+    switch (data.pokemons) {
     | Some(pokemon) => <PokemonList pokemon={pokemon->flattenPokemon} />
     | None => <div> "No Data"->React.string </div>
     }
   | PartialData(data, e) =>
     <div>
-      {switch (data##pokemons) {
+      {switch (data.pokemons) {
        | Some(pokemon) => <PokemonList pokemon={pokemon->flattenPokemon} />
        | None => <div> "No Data"->React.string </div>
        }}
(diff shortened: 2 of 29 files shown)


parkerziegler:
Thanks for this @amiralies! I should be able to get to it this weekend and give a full review. Really appreciate you taking this on!

examples/2-query/src/Container.re
```diff
@@ -11,32 +11,35 @@ module GetAllPokemons = [%graphql
...
+};
 
 [@react.component]
 let make = () => {
   let (Hooks.{response}, _) =
-    Hooks.useQuery(~request, ~requestPolicy=`CacheFirst, ());
+    Hooks.useQuery(
+      ~query=(module GetAllPokemons),
+      ~requestPolicy=`CacheFirst,
+      GetAllPokemons.makeVariables(~first=151, ()),
```
parkerziegler, line 30: Nit – I'd prefer our examples to use record syntax rather than `makeVariables` for supplying variables to hooks:

```suggestion
      GetAllPokemons.{ first: 151 },
```

examples/2-query/src/Pokemon.re
```diff
@@ -20,21 +20,24 @@ module GetPokemon = [%graphql
 
 [@react.component]
 let make = (~pokemon: string) => {
-  let request = GetPokemon.make(~name=pokemon, ());
-  let (Hooks.{response}, _) = Hooks.useQuery(~request, ());
+  let (Hooks.{response}, _) =
+    Hooks.useQuery(
+      ~query=(module GetPokemon),
+      GetPokemon.makeVariables(~name=pokemon, ()),
```
parkerziegler, line 26: +1 to the comment above:

```suggestion
      GetPokemon.{ name: pokemon },
```

examples/2-query/public/index.css
```diff
@@ -13,6 +13,7 @@
   border-radius: 5px;
   box-shadow: 0 10px 15px -3px rgba(0, 0, 0, 0.1),
     0 4px 6px -2px rgba(0, 0, 0, 0.05);
+  margin-top: 15rem;
```
parkerziegler, line 16: Any reason for this? Mind posting a screenshot of the changes? It looks like you altered the structure of the HTML below as well, is there a compelling reason to do this?

examples/3-mutation/src/Grid.re
```diff
@@ -30,34 +18,26 @@ module GetAllDogs = [%graphql
 |}
 ];
 
-let request = GetAllDogs.make();
-
 [@react.component]
 let make = () => {
-  let (Hooks.{response}, _) = Hooks.useQuery(~request, ());
+  let ({Hooks.data}, _) = UseQuery.useQuery(~query=(module GetAllDogs), ());
```
parkerziegler, line 23: I'd really prefer us to use the `response` variant in all examples and always reference hooks from the top-level `Hooks` module. The main reason is to push users towards comprehensively handling error, fetching, and partial data states.

src/Types.re
```diff
@@ -78,14 +78,44 @@ type operation = {
 };
 
 /* The result of the GraphQL operation. */
-type operationResult('data) = {
+type operationResultJs('jsData) = {
```
parkerziegler, line 81: Nit – we prefer to always add Js as a suffix, even on type annotations:

```suggestion
type operationResultJs('dataJs) = {
```

src/Types.re
```diff
@@ -78,14 +78,44 @@ type operation = {
...
 
+/* The result of the GraphQL operation. */
+type operationResult('data) = {
+  data: option('data),
+  error: option(CombinedError.t),
+  extensions: option(Js.Dict.t(string)),
+  response: operationResponse('data),
+  stale: option(bool),
+}
+and operationResponse('data) =
```
parkerziegler, line 97: Is there a benefit to the mutually recursive type here created by the use of `and`? It seems adequate to me to just specify `operationResponse` as its own standalone type and reference it in `response` on L94. I'd only advocate using a mutually recursive type if there was mutual recursion between these types — as far as I can tell, there isn't currently 🤔

src/hooks/UseMutation.re
```diff
@@ -13,130 +13,90 @@ type executeMutation =
...
- * to the GraphQL mutation, and a parse function for decoding the JSON response.
  */
-let useMutation = (~request) => {
-  let query = request##query;
-  let variables = request##variables;
-  let parse = request##parse;
+let useMutation:
+  type data variables.
+    (
+      ~query: (module Types.Operation with
```
parkerziegler, line 40: I'd advocate for one of two options here:

1. Rename this labelled argument to `mutation`.

```suggestion
      ~mutation: (module Types.Operation with
```

2. Don't bother with a labelled argument, since the hook only accepts a single argument.

1 is more compatible with the current API, while 2 maybe makes more sense from a language perspective. What do you think?

src/hooks/UseMutation.re
```diff
@@ -13,130 +13,90 @@ type executeMutation =
...
+                 type t = data and type t_variables = variables)
+    ) =>
+    useMutationResponse(variables, data) =
+  (~query as (module Query)) => {
```
parkerziegler, line 44: Let's name this locally as:

```suggestion
  (~query as (module Mutation)) => {
```

src/hooks/UseSubscription.re
```diff
@@ -71,10 +71,13 @@ external useMemo9:
...
     (
       type acc,
-      type response,
       type ret,
-      ~request: Types.request(response),
-      ~handler: handler(acc, response, ret),
+      type variables,
+      type data,
+      ~query as
+        module Query:
```
parkerziegler, line 78:
```suggestion
      ~subscription as
        module Subscription:
```

parkerziegler (commented):
This looks great @amiralies, a huge, huge thank you for all of this work. A few comments that are mostly small stylistic suggestions, but the changes are very very close. Ping me again when you've addressed the comments, I'll review one last time, and then we can merge this puppy. Thanks again 💯

src/Types.re
amiralies: You're right.

src/hooks/UseMutation.re
amiralies: Both compiles to same js.
I actually like the first approach for consistency between APIs.
AFAIK there shoudn't be any problem with unary function with a (explicitly typed) labeled argument.
you can even omit label (though compiler gives a warning):
```reason
let (state, exec) = Hooks.useMutation((module MyMutation));
```

maybe @jfrolich can also help with this.

examples/2-query/public/index.css
amiralies: Dex is positioned in center of page (notice scroll position):

image

I think correct way to handle this is to use `overflow-y: scroll` on list.

amiralies: I've updated layout using flexbox

image

src/hooks/UseMutation.re
jfrolich: I don't think it makes a difference. If you have multiple arguments it's good to have the query as the first argument because of type inference I think. Otherwise the variable record might not be inferred (not 100% sure).

examples/2-query/src/Container.re
jfrolich: Are you sure a local open is needed here. I don't think it's needed in the Apollo bindings in this scenario.

examples/3-mutation/src/Grid.re
jfrolich: In the new ReScript (8.2+) no annotation is necessary here for the record.

amiralies:
@parkerziegler I think this is ready to be reviewed again :rocket:

parkerziegler (approved):
Beautiful work @amiralies! Thank you again so much for moving this forward!

parkerziegler:
@all-contributors please add @amiralies for code, example.

PR #226, opened 2020-11-07
feat: modify Client API to place variables in last positional argument.

parkerziegler:
This PR does a few things:

1. Updates our `Client` APIs:
  - The `~request` labeled argument is renamed to `~query`, `~mutation`, or `~subscription` depending on the method in use.
  - `variables` are passed as records in the last position, if required. If the query, mutation, or subscription uses no variables, `unit` is passed in the last position.
2. Updates our documentation to reflect the `urql` `peerDependency` introduced in #224 
3. Updates our `Client` documentation to match the latest API changes.
4. Updates our binding to rename `operationName` to `kind` per [this PR](https://github.com/FormidableLabs/urql/pull/1045) in `urql` upstream.

Alongside #225 we're getting _very close_ to a usable `v3`! Moving slowly here but hoping to keep plugging away at everything while I also balance grad school apps this weekend 🤦

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,6 @@
 [![All Contributors](https://img.shields.io/badge/all_contributors-19-orange.svg)](#contributors)
 [![Build Status](https://travis-ci.com/FormidableLabs/reason-urql.svg?branch=master)](https://travis-ci.com/FormidableLabs/reason-urql)
 [![Maintenance Status][maintenance-image]](#maintenance-status)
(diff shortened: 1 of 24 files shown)


kitten (approved):
Nice! I didn't think we'd get to parity this quickly. I really enjoy the `variables` change since it makes natural since. It hasn't been that popular in reason to put an unlabelled required argument last but I've hoped for a long time it'd become more favoured as it's very natural 👏

PR #230, opened 2020-12-07
Use a record to bind CombinedError. Prepare for ReScript migration.

parkerziegler:
With ReScript / `bs-platform` > @8, the syntax for binding JS classes to OCaml classes is no longer supported. In many ways this is a good thing — the old syntax was a bit cumbersome and circuitous. However, we still had our `CombinedError` binding using that syntax, which was always more trouble than it was worth. By the time we receive `CombinedError`, it is an _instance_ not the class itself. Therefore, we can treat it (for our purposes) more or less like a JS object. We can bind this with a record in Reason ✅ 

We also don't need to do any `Js.Nullable` conversion here from what I can tell. Based on [the types of instance properties on `CombinedError`](https://github.com/FormidableLabs/urql/blob/main/packages/core/src/utils/error.ts#L41-L45) we would get `undefined` rather than `null` in all cases for optional properties. These can be safely compiled using `option` in Reason. @kitten let me know if that's not the case and I'll add the appropriate `Js.Nullable` wrapper.

### DIFF
--- a/__tests__/Types_test.re
+++ b/__tests__/Types_test.re
@@ -79,19 +79,14 @@ describe("Types", () => {
             extensions: None,
           };
 
-        let errorJs = {
-          "message": "Error returned by GraphQL API",
-          "graphQLErrors": [|graphQLError|],
-          "networkError": Js.Nullable.null,
-          "response": Js.Nullable.null,
-        };
-
         let error =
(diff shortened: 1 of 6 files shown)


kitten (approved):
Niceeee 🙌

PR #244, opened 2020-12-27
feat(exchanges): add binding for @urql/retry-exchange

parkerziegler:
This PR adds support for using [`@urql/retry-exchange`](https://formidable.com/open-source/urql/docs/advanced/retry-operations/).

The binding for this is fairly thin, although we did need to introduce a `makeRetryExchangeOptions` function to support the creation of a `retryExchangeOptions` record. There are generally two mechanisms for doing this type of "object creation with potentially undefined values" workflow. The first is what we do here, where we have a function that uses optional labelled arguments to generate a record. For example, calling `makeRetryExchaneOptions` with no arguments (just the positional `unit` parameter):

```rescript
open ReasonUrql

let retryExchangeOptions = Client.Exchanges.makeRetryExchangeOptions();
```

results in the following object on the JS-side:

```javascript
{
  initialDelayMs: undefined,
  maxDelayMs: undefined,
  maxNumberAttempts: undefined,
  randomDelay: undefined,
  retryIf: undefined,
}
```

while calling it with _some_ arguments:

```rescript
open ReasonUrql

let retryExchangeOptions =
  Client.Exchanges.makeRetryExchangeOptions(~initialDelayMs=2000, ~randomDelay=false, ())
```

results in the following object on the JS-side:

```javascript
{
  initialDelayMs: 2000,
  maxDelayMs: undefined,
  maxNumberAttempts: undefined,
  randomDelay: false,
  retryIf: undefined,
}
```

We _could_ also use an [abstract record](https://rescript-lang.org/docs/manual/latest/generate-converters-accessors#convert-record-type-to-abstract-record), but these are generally recommended against because of the ergonomics. The one case where they're useful is when compiling an unspecified property explicitly to `undefined` would result in negative behavior. In this case, `urql` checks [the truthiness of all properties and supplies defaults](https://github.com/FormidableLabs/urql/blob/main/exchanges/retry/src/retryExchange.ts#L37-L43) downstream, so we should be safe explicitly compiling to `undefined`!


kitten (approved):
I'm a bit concerned about adding more snapshots that save the Client's shape 🤔 other than that LGTM of course ✌️

parkerziegler:
Good shout @kitten. I haven't figured out a good way of testing these new exchanges TBH. I think the tests of "can they be applied" actually aren't meaningful, because even a Client snapshot doesn't verify that the exchange is being applied. We'd need more sophisticated logic to execute requests with a Client instance and spy on some behavior of the exchange. That's more effort than I want to invest just yet, but I'll remove the unnecessary snapshots here 👍

PR #248, opened 2021-01-12
Add rescript keyword to package.json

Kingdutch:
Since this package is compatible with ReScript it makes sense to 
add the `rescript` keyword to aid in discoverability and cause the 
package to be picked up by [the package overview on rescript-lang.org](https://rescript-lang.org/packages).

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   "keywords": [
     "reason",
     "reasonml",
+    "rescript",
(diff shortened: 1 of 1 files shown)


parkerziegler (approved):

parkerziegler:
@all-contributors please add @Kingdutch for code.

Issue #249, opened 2021-01-12
Rename package to rescript-urql


amiralies:
Does it make sense to rename the package to `rescript-urql`?

parkerziegler:
Hey @amiralies!

Yes, I do think eventually we'll want to rename the repository and package. But I think there are a few steps we'll need to plan out:

- Secure `npm` package name.
- Introduce a notification to the README instructing people to migrate.
- Do the switch as a major version bump (4.0.0).

I also need to check internally with our OSS Lead at Formidable to make sure we coordinate this process internally, update our website, etc.

So, yes, we'll do this, but I just need to plan out a cycle for it!

PR #254, opened 2021-03-07
feat: add support for @rescript/react, urql 2, and bs-platform 9

### LINKED ISSUE #253: Using @rescript/react
Is there a way to use `@rescript/react` instead of `reason-react` with this package? Not having `reason-react` installed causes bsb to emit errors after installing this PPX, it appears that it has `reason-react` as an (undocumented) hard requirement.

Side note, it would be super nice to be able to get bindings for just `@urql/core`.

Thanks for your work! I use the PPX in a different project that uses `reason-react` and it's absolutely wonderful.

parkerziegler:
Fix #253 

This PR begins the process of migrating us to a whole slate of new dependencies:

- [`@rescript/react`](https://github.com/rescript-lang/rescript-react) — this is where the ReScript core team will be maintaining React moving forward.
- [`urql`@2.0.1](https://github.com/FormidableLabs/urql/blob/main/packages/react-urql/CHANGELOG.md#urql) — this needs  no explanation 😅 
- [`bs-platform`@9.0.1](https://rescript-lang.org/blog/release-9-0) — this just picks up the new compiler version and, fortunately, we have no breaking code!

I also removed the `pollInterval` option per the `urql`@2.0.0 CHANGELOG and updated a few internals to support this.

The next step in this migration is to chat about a new publishing scheme for this package. In #249, there was some discussion about renaming the project `rescript-urql`. I support this change with a few suggestions:

- Project renamed to `rescript-urql`.
- Package published  as `@urql/rescript` — this better aligns with our publishing scopes for `vue`, `svelte`, `preact`, etc. even though this package still lives separately in this repo.

I'd be curious for thoughts / feelings on the above, and also advice on what our deprecation strategy for the `reason-urql` namespace should be.

Lastly, there was some discussion in #253 about separating out bindings for `@urql/core` so that ReScript users can access _just those_. I like the sound of this change, but I'm not fully sure about the scope of what that looks like ("Hello monorepo?"). So, some food for thought!


parkerziegler:
Thanks for the heads up @gaku-sei! I'll wait until v9.0.2 lands. I am planning on also renaming this package to `@urql/rescript` for a v4.0.0 (and to bring it more consistently under the family of `urql` packages). So this won't be out and available for a little bit! Thanks for testing the branch early though!

package.json
```diff
@@ -35,22 +35,22 @@
...
-    "bs-platform": "^8.3.2",
-    "gentype": "^3.38.0",
+    "bs-platform": "^9.0.1",
+    "gentype": "^3.44.0",
     "graphql": "^15.4.0",
     "npm-run-all": "^4.1.5",
     "react": "^16.8.0",
-    "reason-react": "^0.9.1",
-    "urql": "~1.11.2"
+    "urql": "~2.0.1"
```
kitten, line 46: I'd love for this to switch over to a simple minor-range, since future changes shouldn't affect the bindings. I think the patch range was rather temporary, if I'm not mistaken, to prevent other changes from interferring with the bindings slowly catching up, so since we're all caught up now it should be safe to bump this to the minor range

package.json
```diff
@@ -35,22 +35,22 @@
...
     "@reasonml-community/graphql-ppx": "^1.0.1",
+    "@rescript/react": "^0.10.1",
     "graphql": "^0.11.0 || ^0.12.0 || ^0.13.0 || ^14.0.0 || ^15.0.0",
     "react": "^16.8.0",
-    "reason-react": "^0.7.0",
-    "urql": "~1.11.0"
+    "urql": "~2.0.0"
   },
   "resolutions": {
     "wonka": "5.0.0-rc.1"
```
kitten, line 56: I suppose, this is still a permanent fix that all users need to add, right? Let's chat about how we can cleanly migrate all of urql to a new version, since I kind of got stuck on updating everything

package.json
parkerziegler: Yep, you're right! I agree that a more permissive minor range sounds appropriate!

package.json
parkerziegler: Sure thing — yeah the internals of how ReScript compiles variants really :rekt: us. I'll see if I can snag some of your time later in March (on PTO next week) to discuss what an upgrade strategy might look like.

